PR #2, opened 2015-11-02
Integrated Panel Manager into View Manager

liana-yang:
Added a panel manager to the system, which would interact with the view manager.

Confusions:
1. Panel manager functions are called directly after the view manager functions. There are also other ways to call panel manager like using callbacks or trigger events. But I am not sure which one is better because every part is written as an independent global variant and I think direct call is an efficient way. Do you have any other ideas about the advantages of other ways?
2. The way of initialization of most components is .load(). .load() is an Ajax function and executes asynchronously. Calling components in default (like execute Test.run() in core.init()) would caused async execution problems and I got confused about the reason why using this kind of setting.

Next steps:
1. Fix the bugs of components interaction.
2. Improve some interaction methods.


yubowenok:
1. My idea is to call a panel popup function from the view object when a view is activated. The panel popup function returns the view the container of the panel. When the view is closed, the view itself calls a view close handler of the panel manager. PanelManager may be globally visible for this purpose.
2. Data loading has to be async as the server needs time to read and process the data. Currently Core is only inited when the page is ready (see index.html). Then the view creations are made. Each view's data is loaded asyncronously. When the data is complete, a "genotet.loadComplete" event is triggered so that the view starts to render.

css/panel-manager-style.css
```diff
@@ -0,0 +1,50 @@
...
+}
+#tab-slidebar {
+    padding-left: 0px;
+}
+.sideways {
+    padding-top: 0px;
+}
+#tab-content-slidebar {
+    margin-top: 10px;
+}
```
yubowenok: Please add a new line at the end of each file to get rid of this :)

js/panel-manager.js
```diff
@@ -0,0 +1,57 @@
+// var count = 0;
+
+var PanelManager = {
+//$(document).ready(function(){
+//  $('#btn-add').click(function(e){
+  init: function(e){
+    $('.panel-manager').load('templates/panel-manager.html', function(e){
+      $('#slidebar').on('click', function(e){
+        $('#slider').toggle('slide', { direction: 'right' }, 300);
```
yubowenok: Please use constants for the constant numbers, e.g. "PanelManager.TRANSITION_TIME". Looks like 300 is used repeatedly.

js/panel-manager.js
```diff
@@ -0,0 +1,57 @@
...
+          $('#icon-button').toggleClass('glyphicon-chevron-right glyphicon-chevron-left');
+        }, 300);
+      });
+    });
+  },
+
+  addPanel: function(view){
+//     count++;
+    var count = $('.sideways').children().length;
+//    e.preventDefault();
```
yubowenok: Please cleanup the commented code a bit. I'd recommend using bulk comments /_..._/ and add "TODO(liana): ..." to describe why the commented code is left. If the code will no longer be needed, then remove the commented code.

js/panel-manager.js
```diff
@@ -0,0 +1,57 @@
...
+//    $('.sideways').append('<li id="li'+count+'"><a href="#view'+count+'" data-toggle="tab">view'+count+'</a></li>');
+//    $('.sideways').append('<li id="li'+count+'"><a href="#view'+count+'" data-toggle="tab">'+viewName+'</a></li>');
+    var viewID = view.name().replace(/\s/g, '');
+    $('.sideways').append('<li id="li'+viewID+'"><a href="#'+viewID+'" data-toggle="tab">'+view.name()+'</a></li>');
```
yubowenok: It's a bit nasty to hard code DOM with jquery like this (sorry about my bad previous practice! :P ). Is there a better way to add DOM element? 

If not, please format the line a bit. In JS, we add a single space between the operands, e.g.
'string text...' + viewID + 'string text...'.
You may use multiple lines when there are string concats.

js/panel-manager.js
```diff
@@ -0,0 +1,57 @@
...
+    if ($('#slidebar').css('right') == '0px'){ $('#slidebar').trigger('click'); }
+  },
+//  $('#btn-remove').click(function(e){
+//  $.when($.ajax(ViewManager.closeView())).done(function(e){
+  removePanel: function(viewName){
+    var count = $('.sideways').children().length - 1;
+    var actived = false;
+//    if ($('#li'+count).hasClass('active') == true){ actived = true; }
+    var viewID = viewName.replace(/\s/g, '');
+    if ($('#li'+viewID).hasClass('active') == true){ actived = true; }
```
yubowenok: Always use 3 lines for an "if" block.
if (condition) {
  // code ...
}

"== true" can be omitted here.

js/panel-manager.js
```diff
@@ -0,0 +1,57 @@
...
+    $('.tab-content').append('<div class="tab-pane" id="'+viewID+'">property'+count+'</div>');
+    $('#li'+viewID+' a[href="#'+viewID+'"]').tab('show');
+    $('.panel-manager').css('display','inline');
+    if ($('#slidebar').css('right') == '0px'){ $('#slidebar').trigger('click'); }
+  },
+//  $('#btn-remove').click(function(e){
+//  $.when($.ajax(ViewManager.closeView())).done(function(e){
+  removePanel: function(viewName){
+    var count = $('.sideways').children().length - 1;
+    var actived = false;
```
yubowenok: Do you mean "active" or "activated"?

templates/panel-manager.html
```diff
@@ -0,0 +1,42 @@
+<!--<!DOCTYPE html>-->
```
yubowenok: Please remove unused code :)

js/panel-manager.js
```diff
@@ -0,0 +1,57 @@
...
+//    if ($('#li'+count).hasClass('active') == true){ actived = true; }
+    var viewID = viewName.replace(/\s/g, '');
+    if ($('#li'+viewID).hasClass('active') == true){ actived = true; }
+//    $('#li'+count).remove();
+    $('#li'+viewID).remove();
+    $('#'+viewID).remove();
+    count--;
+    if (actived == true){ $('#li'+viewID+' a[href="#'+viewID+'"]').tab('show'); }
```
yubowenok: ditto

css/panel-manager-style.css
```diff
@@ -0,0 +1,50 @@
...
+    top: 30px;
+    height: auto;
+    width: 200px;
+    overflow-y: scroll;
+    position: fixed;
+}*/
+.panel-manager {
+    display: none;
+}
+#slidebar {
```
yubowenok: Please add a blank line between each css block.

css/panel-manager-style.css
```diff
@@ -0,0 +1,50 @@
+/*.panel-manager {
```
yubowenok: Looks like this file is 4-indented?

css/panel-manager-style.css
```diff
@@ -0,0 +1,50 @@
...
+    /*transform: rotate(-90deg);*/
+}
+#slidebar button {
+    border-radius: 4px 0 0 4px;
+    border-right: none;
+    height: 123px;
+    width: 30px;
+    /*border-right: none;*/
+}
+#slider {
```
yubowenok: The id "slider" might be a bit too common and risky to use here. Is there a more specfic name for it?

css/panel-manager-style.css
liana-yang: Fixed.

js/panel-manager.js
liana-yang: Fixed.

js/panel-manager.js
liana-yang: Fixed.

js/panel-manager.js
liana-yang: Fixed.

js/panel-manager.js
liana-yang: Fixed.

js/panel-manager.js
liana-yang: Fixed.

css/panel-manager-style.css
liana-yang: Fixed.

css/panel-manager-style.css
liana-yang: Fixed.

templates/panel-manager.html
liana-yang: Fixed.

css/bootstrap.vertical-tabs.min.css
```diff
@@ -0,0 +1,9 @@
+/*!
```
yubowenok: Maybe use bower to install this?

css/panel-manager-style.css
```diff
@@ -0,0 +1,48 @@
...
+
+#slidebar button {
+  border-radius: 4px 0 0 4px;
+  border-right: none;
+  height: 123px;
+  width: 30px;
+  /* border-right: none; */
+}
+
+#slider {
```
yubowenok: The name slider is a bit too pervasive here. Maybe change to sth more specific such as sidepanel-slider.

liana-yang: Fixed.

js/panel-manager.js
liana-yang: Fixed.

css/panel-manager-style.css
liana-yang: Fixed.

css/bootstrap.vertical-tabs.min.css
liana-yang: Fixed.

index.html
```diff
@@ -122,13 +124,14 @@
...
-  <div id="main" class="container-fluid">
-  </div>
+  <div id="main" class="container-fluid"></div>
   <div id="modal" class="modal fade">
     <div class="modal-dialog">
       <div class="modal-content">
       </div>
     </div>
   </div>
+  <div class="panel-manager">
```
yubowenok: Umm looks like panel-manager is a controller class in code and this could be "side-panel" instead...

css/panel-manager-style.css
```diff
@@ -0,0 +1,49 @@
+.panel-manager {
```
yubowenok: This file can be renamed as "panel-manager.css" or "side-panel.css".

css/panel-manager-style.css
```diff
@@ -0,0 +1,49 @@
...
+
+#tab-slider {
+  padding-left: 0px;
+}
+
+.sideways li:first-child {
+  display: none;
+}
+
+.sideways li:nth-child(2) {
```
yubowenok: What is the specialty of the second child?

js/panel-manager.js
```diff
@@ -0,0 +1,56 @@
+var PanelManager = {
+  TRANSITION_TIME: 300,
+
+  init: function(e) {
+    $('.panel-manager').load('templates/panel-manager.html', function(e) {
+      $('#btnToggle').on('click', function(e) {
+        $('#slider').animate({
+          'right' : $('#slider').css('right') == '0px' ? '-208px' : '0px'
```
yubowenok: Could you make '208px' a constant too? BTW 'px' may be omitted if I recall correctly...

js/panel-manager.js
```diff
@@ -0,0 +1,56 @@
...
+        setTimeout(function(e) {
+          $('#icon-button').toggleClass('glyphicon-chevron-right glyphicon-chevron-left');
+        }, PanelManager.TRANSITION_TIME);
+      });
+    });
+  },
+
+  addPanel: function(viewName) {
+    var viewID = viewName.replace(/\s/g, '');
+    $('#li-init').clone().attr('id', 'li-' + viewID).appendTo('.sideways');
```
yubowenok: Please format the jQuery calls in a multi-line chain style.

Are these ids for the panels? If so please rename them to '#panel-viewID'.

index.html
liana-yang: Fixed.

css/panel-manager-style.css
liana-yang: Fixed.

css/panel-manager-style.css
liana-yang: Because there exists a init item which is used to be cloned, the first panel we append would become the second child. There are some css style need to be revised for the first panel. That's why I call it by the second child.

js/panel-manager.js
liana-yang: Fixed.

js/panel-manager.js
liana-yang: Since tab and view both have its own id, how about name them 'panel-tab-viewID' and 'panel-view-viewID'?

js/side-panel.js
```diff
@@ -0,0 +1,87 @@
...
+  },
+
+  /**
+   * Closes the given panel.
+   * @param {string} viewName Name of the view.
+   */
+  removePanel: function(viewName) {
+    var count = $('.sideways')
+                    .children()
+                        .length - 1;
```
yubowenok: var count = $('.sideways').children().length - 1;

js/side-panel.js
```diff
@@ -0,0 +1,87 @@
...
+    var container = $('#panel-view-' + viewID);
+    return container;
+  },
+
+  /**
+   * Closes the given panel.
+   * @param {string} viewName Name of the view.
+   */
+  removePanel: function(viewName) {
+    var count = $('.sideways')
```
yubowenok: ditto

js/side-panel.js
```diff
@@ -0,0 +1,87 @@
+var PanelManager = {
+  TRANSITION_TIME: 300,
+  TOGGLE_DISTANCE: 218,
+
+  init: function(e) {
+    $('.side-panel').load('templates/side-panel.html', function(e) {
+      $('#btnToggle').on('click', function(e) {
+        $('#slider').animate({
+          'right' : $('#slider').css('right') == '0px' ? - PanelManager.TOGGLE_DISTANCE + 'px': '0px'
```
yubowenok: You may omit the single quotes for 'right'.

js/side-panel.js
liana-yang: Fixed.

js/side-panel.js
liana-yang: The single quotes cannot be removed according to http://api.jquery.com/css/

js/side-panel.js
Shauro: Fixed.

js/side-panel.js
liana-yang: There will render error "Uncaught ReferenceError: right is not defined" when the single quotes are removed.

PR #22, opened 2015-11-12
Rendered expression matrix as static heatmap and gene labels.

liana-yang:
I rendered expression matrix as static heatmap and gene labels. I didn't finish the condition labels and and computing the horizontal/vertical text margin allocated for the heatmap labels because the interactive part and zooming part is not rendered yet. I am not very clear about the position detail of the expression view (including the zooming setting). Besides, I think the color scale need to be revised because it is not very easy to read now.


components/expression/expression-renderer.js
```diff
@@ -179,16 +179,31 @@ ExpressionRenderer.prototype.drawMatrixCells_ = function() {
     width: this.heatmapWidth_,
     height: this.heatmapHeight_
   }]);
-  cells.enter().append('rect')
-    .classed('cell', true);
+  var heatmapData = this.dataReady_();
```
yubowenok: this.dataReady_() returns boolean. Looks like I missed the base class documentation.

Use "this.data.matrix" to get the data object.

components/expression/expression-renderer.js
```diff
@@ -179,16 +179,31 @@ ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+  console.log(heatmapData);
+  var colorScale = d3.scale.linear()
```
yubowenok: Data.redYelloScale

components/expression/expression-renderer.js
```diff
@@ -179,16 +179,31 @@ ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+      .domain([heatmapData.valueMin, heatmapData.valueMax])
+      .range(['white', 'red']);
+  var heatmapRow = cells.data(heatmapData.values)
+      .enter().append('g')
+      .attr("transform", "translate(" + 30 + "," + 10 + ")");
```
yubowenok: The rows will need exit() and update too.

components/expression/expression-renderer.js
```diff
@@ -179,16 +179,31 @@ ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+  var heatmapRects = heatmapRow.selectAll('.rect')
+      .data(function(d) { return d; })
+      .enter().append('rect')
+      .attr('width', 5)
```
yubowenok: width and height shall be determined by the current heatmap size, in order to fill up the entire heatmap area.

components/expression/expression-renderer.js
```diff
@@ -179,16 +179,31 @@ ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+  var heatmapData = this.dataReady_();
+  console.log(heatmapData);
+  var colorScale = d3.scale.linear()
+      .domain([heatmapData.valueMin, heatmapData.valueMax])
+      .range(['white', 'red']);
+  var heatmapRow = cells.data(heatmapData.values)
+      .enter().append('g')
+      .attr("transform", "translate(" + 30 + "," + 10 + ")");
+  var heatmapRects = heatmapRow.selectAll('.rect')
+      .data(function(d) { return d; })
```
yubowenok: Use "_.identity".

components/expression/expression-renderer.js
```diff
@@ -179,16 +179,31 @@ ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+      .enter().append('rect')
+      .attr('width', 5)
+      .attr('height', 15)
+      .attr('x', function(d, i, j) { return i * 5 })
+      .attr('y', function(d, i, j) { return j * 15 })
+      .style('fill', function(d) { return colorScale(d) })
```
yubowenok: missing semicolon

components/expression/expression-renderer.js
```diff
@@ -179,16 +179,31 @@ ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+      .range(['white', 'red']);
+  var heatmapRow = cells.data(heatmapData.values)
+      .enter().append('g')
+      .attr("transform", "translate(" + 30 + "," + 10 + ")");
+  var heatmapRects = heatmapRow.selectAll('.rect')
+      .data(function(d) { return d; })
+      .enter().append('rect')
+      .attr('width', 5)
+      .attr('height', 15)
+      .attr('x', function(d, i, j) { return i * 5 })
```
yubowenok: return on another line

components/expression/expression-renderer.js
```diff
@@ -201,17 +216,24 @@ ExpressionRenderer.prototype.drawMatrixGeneLabels_ = function() {
...
-  }]);
+  //var labels = this.svgGeneLabels_.selectAll('text').data([{
+  //  label: 'a horizontal label'
+  //}]);
+  var geneLabelsData = this.dataReady_().geneNames;
+  var labels = this.svgGeneLabels_.selectAll('text').data(geneLabelsData);
   labels.enter().append('text')
-    .classed('gene-label', true);
+      .text(function(d) { return d; })
+      .attr('x', 0)
```
yubowenok: x, y need to be set in update, instead of enter(). Otherwise this will not handle redraw (e.g. when view is resized).

components/expression/expression-renderer.js
```diff
@@ -201,17 +216,24 @@ ExpressionRenderer.prototype.drawMatrixGeneLabels_ = function() {
...
+      .attr('y', function(d, i) { return i * 15 })
+      .style('text-anchor','middle')
+      .attr("transform", "translate(" + 30 + "," + 20 + ")")
```
yubowenok: Why the magic number [30, 20]? They were placeholders...

Use Utils.getTransform().

components/expression/expression-renderer.js
```diff
@@ -179,16 +179,31 @@ ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
   cells.exit().remove();
   cells
-    .attr('width', function(cell) {
-      return cell.width;
-    })
-    .attr('height', function(cell) {
-      return cell.height;
-    });
+      .attr('width', function(cell) {
+        return cell.width;
```
yubowenok: 2-indent...

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
Shauro: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
Shauro: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
```diff
@@ -67,16 +67,16 @@ ExpressionRenderer.prototype.DEFAULT_PROFILE_HEIGHT = 300;
  this.heatmapMargin = 3;
  this.heatmapFontSize = 8.5;
  this.heatmapFontHSize = 8.5;
-*/
+ */
 
 
 /** @inheritDoc */
-ExpressionRenderer.prototype.init = function() {
+ExpressionRenderer.prototype.init = function () {
```
yubowenok: There should be NO space between 'function' and '('

components/expression/expression-renderer.js
```diff
@@ -113,19 +113,19 @@ ExpressionRenderer.prototype.initLayout = function() {
 };
 
 /** @inheritDoc */
-ExpressionRenderer.prototype.layout = function() {
+ExpressionRenderer.prototype.layout = function () {
   // Gets the label sizes so as to set the offset of heatmap SVG.
   this.getHeatmapLabelSizes_();
   // Compute the shifting sizes.
   this.profileHeight_ = this.data.options.showProfiles ?
-      this.DEFAULT_PROFILE_HEIGHT : 0;
```
yubowenok, line 119: If the line continues from the previous line, and breaks at an operator, we use 4 indent instead of 2.

components/expression/expression-renderer.js
```diff
@@ -195,53 +222,79 @@ ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+  var heatmapData = this.data.matrix;
+  var conditionLabelsData = heatmapData.conditionNames;
+  var cellWidth = this.heatmapWidth_ / conditionLabelsData.length;
+  //var labels = this.svgConditionLabels_.selectAll('text').data([{
+  //  label: 'a vertical label'
+  //}]);
+  var labels = this.svgConditionLabels_.selectAll('text').data(conditionLabelsData);
   labels.enter().append('text')
+    .text(_.identity)
+    .attr('x', 0)
```
yubowenok: setting x to zero can be omitted I think

components/expression/expression-renderer.js
```diff
@@ -195,53 +222,79 @@ ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+    .style('text-anchor', 'right')
     .classed('condition-label', true);
   labels.exit().remove();
   labels
-    .attr('transform', Utils.getTransform([0, 0], 1, 90))
-    .text(function(condition) {
-      return condition.label
-    });
+    .attr('transform', Utils.getTransform([this.geneLabelWidth_, this.conditionLabelHeight_], 1, -90))
```
yubowenok: line too long

components/expression/expression-renderer.js
```diff
@@ -265,161 +318,161 @@ ExpressionRenderer.prototype.getHeatmapLabelSizes_ = function() {
     this.geneLabelWidth_ = 30;
   }
   if (!this.data.options.showConditionLabels) {
-    this.conditionLabelHeight_ = 0;
+    this.conditionLabelHeight_ = 100;
```
yubowenok: This should be zero. If showConditionLabels option is set to false, then conditionLabelHeight_ must be ZERO.

components/expression/expression-renderer.js
```diff
@@ -265,161 +318,161 @@ ExpressionRenderer.prototype.getHeatmapLabelSizes_ = function() {
...
   } else {
     // To implement... replace dummy value.
-    this.conditionLabelHeight_ = 50;
+    this.conditionLabelHeight_ = 200;
```
yubowenok: This value should be computed based on the longest label, and not a constant.

components/base/view.js
```diff
@@ -143,11 +146,13 @@ genotet.View.prototype.headerText = function(headerText) {
 /**
  * Makes the view appear focused.
```
yubowenok, line 150: Please add document for _sendSignal_.

components/expression/expression-panel.js
```diff
@@ -93,6 +93,32 @@ genotet.ExpressionPanel.prototype.updateGenes = function(genes) {
   });
 };
```
yubowenok: Missing documents for all the functions below.

components/expression/expression-panel.js
```diff
@@ -93,6 +93,32 @@ genotet.ExpressionPanel.prototype.updateGenes = function(genes) {
...
 
+genotet.ExpressionPanel.prototype.setCellInfo_ = function(geneName, conditionName, container) {
+  container.html(this.container_.find('#cell-info-template').html());
+  container.children('#gene').children('span')
+    .text(geneName);
+  container.children('#condition').children('span')
+    .text(conditionName);
+};
+
+genotet.ExpressionPanel.prototype.hideEdgeInfo_ = function() {
```
yubowenok: This function is misnamed.

components/expression/expression-renderer.js
```diff
@@ -22,13 +22,43 @@ genotet.ExpressionRenderer = function(container, data) {
...
    * This value will be zero when condition labels are not shown.
    * @private {number}
    */
   this.conditionLabelHeight_ = 0;
 
   /**
+   * The factor of the maximum height of the vertical experiment condition labels.
+   * @private {number}
+   */
+  this.conditionLabelHeightFactor_ = 6.501175;
```
yubowenok: If this is a constant, then it should be named as CONDITION_LABEL_HEIGHT_FACTOR_ and be an attribute of the class prototype. Same for the constants below.

components/expression/expression-renderer.js
```diff
@@ -172,21 +211,66 @@ genotet.ExpressionRenderer.prototype.drawExpressionMatrix_ = function() {
...
-  var cells = this.svgHeatmapContent_.selectAll('rect.cell').data([{
-    width: this.heatmapWidth_,
-    height: this.heatmapHeight_
-  }]);
-  cells.enter().append('rect')
-    .classed('cell', true);
+  var heatmapData = this.data.matrix;
+  var cellWidth = this.heatmapWidth_ / heatmapData.conditionNames.length;
+  var cellHeight = this.heatmapHeight_ / heatmapData.geneNames.length;
+  //console.log(heatmapData);
```
yubowenok: Please delete.

components/expression/expression-renderer.js
```diff
@@ -172,21 +211,66 @@ genotet.ExpressionRenderer.prototype.drawExpressionMatrix_ = function() {
...
+  var colorScale = d3.scale.linear()
+    .domain([heatmapData.valueMin, heatmapData.valueMax])
+    .range(genotet.data.redYellowScale);
+
+  this.svgHeatmapContent_.selectAll('g').remove();
```
yubowenok: Why removing all 'g'? This means you want to re-render everything again, and this should be avoided.

For example, when the view is resized, you only need to adjust the cell sizes. But if you remove all g's, then every element has to be added to the HTML again, which is costly. By doing so, you miss the purpose of d3.selectAll().data().

components/expression/expression-renderer.js
```diff
@@ -172,21 +211,66 @@ genotet.ExpressionRenderer.prototype.drawExpressionMatrix_ = function() {
...
   cells.exit().remove();
-  cells
-    .attr('width', function(cell) {
-      return cell.width;
+  var heatmapRects = heatmapRows.selectAll('.rect')
+    .data(_.identity)
+    .enter().append('rect')
+    .attr('width', cellWidth)
+    .attr('height', cellHeight)
+    .attr('x', function(d, i, j) {
```
yubowenok: Please remove the trailing function parameter that is not used. Please rename 'd' so that it is more explicit what the binded data is.

components/expression/expression-renderer.js
```diff
@@ -172,21 +211,66 @@ genotet.ExpressionRenderer.prototype.drawExpressionMatrix_ = function() {
...
+      this.conditionLabelHeight_ + this.profileHeight_
+    ]));
   cells.exit().remove();
-  cells
-    .attr('width', function(cell) {
-      return cell.width;
+  var heatmapRects = heatmapRows.selectAll('.rect')
+    .data(_.identity)
+    .enter().append('rect')
+    .attr('width', cellWidth)
```
yubowenok, line 326: You are only setting properties on enter(). This works because you are re-rendering everything.
1) Please fix your rendering scheme so that it handles enter(), exit() and update properly.
2) Keep all properties NOT subject to change when re-rendered in enter(). Move those properties that would change to update.

components/expression/expression-renderer.js
```diff
@@ -172,21 +211,66 @@ genotet.ExpressionRenderer.prototype.drawExpressionMatrix_ = function() {
...
     })
-    .attr('height', function(cell) {
-      return cell.height;
-    });
+    .on('mouseover', function(d, i, j) {
+      var hoverCell = d3.event.target;
+      this.signal(
+        'cellHover',
+        [
+          hoverCell,
```
yubowenok: Create an object for this and name the attributes. Otherwise you will have to do data[0], data[1] in signal handler, which is confusing.

components/expression/expression-renderer.js
```diff
@@ -198,18 +282,24 @@ genotet.ExpressionRenderer.prototype.drawMatrixGeneLabels_ = function() {
...
-  }]);
+  var heatmapData = this.data.matrix;
+  var geneLabelsData = heatmapData.geneNames;
+  var cellHeight = this.heatmapHeight_ / geneLabelsData.length;
+  var labels = this.svgGeneLabels_.selectAll('text').data(geneLabelsData);
   labels.enter().append('text')
+    .attr("transform", genotet.utils.getTransform([
+      this.geneLabelWidth_ - this.labelMargin_,
+      this.conditionLabelHeight_ + this.profileHeight_ + this.topMargin_ + cellHeight / 2
+    ], 1))
```
yubowenok: The scale ", 1" can be omitted.

components/expression/expression-renderer.js
```diff
@@ -221,18 +311,24 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
-    label: 'a vertical label'
-  }]);
+  var heatmapData = this.data.matrix;
+  var conditionLabelsData = heatmapData.conditionNames;
+  var cellWidth = this.heatmapWidth_ / conditionLabelsData.length;
+  var labels = this.svgConditionLabels_.selectAll('text').data(conditionLabelsData);
   labels.enter().append('text')
+    .attr('transform', genotet.utils.getTransform([
+      this.textHeight_ / 2 + cellWidth / 2,
+      this.conditionLabelHeight_ + this.profileHeight_ - this.labelMargin_
```
yubowenok: If you are adding margin to every label, then you may want to translate the label group (which contains all labels). You may also consider rotating the label group.

components/expression/expression-renderer.js
```diff
@@ -254,19 +350,29 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
-    // To implement... replace dummy value.
-    this.geneLabelWidth_ = 30;
+  var heatmapData = this.data.matrix;
+  var geneLabelsData = heatmapData.geneNames;
+  var conditionLabelsData = heatmapData.conditionNames;
+  this.geneLabelWidth_ = 0;
+  this.conditionLabelHeight_ = 0;
+
+  if (this.data.options.showGeneLabels) {
+    for (var i = 0; i < geneLabelsData.length; i++) {
```
yubowenok: Use Array.forEach to iterate javascript arrays.

Probably you can do:

``` js
this.geneLabelWidth_ = Math.max.apply(this, geneLabelsData.map(function(s) {
  return s.length; 
}));
```

Or

``` js
this.geneLabelWidth_ = d3.max(geneLabelsData.map(function(s) {
  return s.length;
});
```

components/expression/expression-renderer.js
```diff
@@ -254,19 +350,29 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
-    // To implement... replace dummy value.
-    this.conditionLabelHeight_ = 50;
+  if (this.data.options.showConditionLabels) {
+    for (var i = 0; i < conditionLabelsData.length; i++) {
+      if (conditionLabelsData[i].length > this.conditionLabelHeight_) {
+        this.conditionLabelHeight_ = conditionLabelsData[i].length;
+      }
+    }
+    this.conditionLabelHeight_ *= this.conditionLabelHeightFactor_;
+    this.conditionLabelHeight_ += this.labelMargin_;
```
yubowenok: Do not add margin to label height. LabelHeight is defined as the height of the label text itself, not including the margins.

components/expression/expression.css
```diff
@@ -1,3 +1,9 @@
+.panel.expression .box {
```
yubowenok: Maybe move this to panel.css.

``` css
.panel .box {
}
```

components/expression/expression.js
```diff
@@ -30,6 +30,24 @@ genotet.ExpressionView = function(viewName, params) {
   $(this.container).on('genotet.ready', function() {
     this.loader.load(params.matrixName, params.geneRegex, params.condRegex);
   }.bind(this));
+
+  $(this.renderer)
+    .on('genotet.cellHover', function(event, data) {
+      var cellSelection = d3.select(data[0]);
+      cellSelection.style('stroke', 'white');
```
yubowenok: Please do not style here. Style is the job of renderer, so the code needs to be in renderer.js instead.

js/panel-manager.js
```diff
@@ -69,33 +75,62 @@ genotet.panelManager.togglePanel_ = function() {
...
+  $('#' + tabContentID).addClass('active');
+},
+
+/**
  * Creates a panel with the given name.
  * @param {string} viewName Name of the view.
  */
 genotet.panelManager.addPanel = function(view) {
-  var viewID = view.name().replace(/\s/g, '');
+  var viewID = view.name().replace(/\s/g, '-');
```
yubowenok: It is bad code to do such replace every time here. You may want to create a map that stores (originalName, renamedName).

However, I would recommend you using a map from view.name() to your tab/panel jQuery object. By doing so you can get rid of the name replace magic.

js/panel-manager.js
```diff
@@ -69,33 +75,62 @@ genotet.panelManager.togglePanel_ = function() {
...
+  $('.sideways li a').off().click(function(event) {
+    if (!this.showPanel_) {
+      this.togglePanel_();
+    }
+    event.stopPropagation();
+    var clickedTabID = $(event.target).parent().attr('id');
+    var clickedViewID = clickedTabID.replace('panel-tab-', '');
+    var clickedViewName = clickedViewID.replace(/\-/g, ' ');
+    var clickedView = genotet.viewManager.views[clickedViewName];
+    $('#main div.focused').removeClass('focused');
```
yubowenok: Why not using viewManager.blurAllViews()?

js/panel-manager.js
```diff
@@ -118,7 +153,7 @@ genotet.panelManager.removePanel = function(viewName) {
     // One is the hidden template to be cloned.
     this.container_.hide();
   }
-};
+},
```
yubowenok: ?

js/panel-manager.js
```diff
@@ -17,6 +17,12 @@ genotet.panelManager.TRANSITION_TIME = 300;
 genotet.panelManager.showPanel_ = true;
 
 /**
+ * Whether sends focused signal..
+ * @private {boolean}
+ */
+genotet.panelManager.sendSignal_ = false;
```
yubowenok: What is this for? You seems to be using only 'false'.

js/panel-manager.js
```diff
@@ -69,33 +75,62 @@ genotet.panelManager.togglePanel_ = function() {
...
-    .attr('id', 'panel-view-' + viewID)
+    .attr('id', tabContentID)
     .appendTo('.tab-content');
-  $('#' + tabID + ' a').trigger('click');
-
+  $(view).on('genotet.focus', function() {
+    var clickedViewID = view.name().replace(/\s/g, '-');
+    this.activatePanel_(clickedViewID);
+  }.bind(this));
+  $('.sideways li a').off().click(function(event) {
```
yubowenok, line 107: Why off()? Looks tricky. If you intentionally do so please add some comments.

js/panel-manager.js
```diff
@@ -69,33 +75,62 @@ genotet.panelManager.togglePanel_ = function() {
...
+    if (!this.showPanel_) {
+      this.togglePanel_();
+    }
+    event.stopPropagation();
```
yubowenok: Please always have stopPropagation()  at the top or end of a handler.

components/expression/expression-renderer.js
liana-yang, line 326: Need discussion.

components/expression/expression-renderer.js
liana-yang: Need discussion.

components/expression/expression-renderer.js
liana-yang: Need discussion.

components/base/view.js
liana-yang, line 150: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

js/panel-manager.js
liana-yang: True is used in view.js to avoid cyclic signal between focusing view and activating panel.

js/panel-manager.js
liana-yang, line 107: Remove the click event handler to avoid multiple executions.

components/expression/expression-loader.js
```diff
@@ -52,117 +54,237 @@ genotet.ExpressionLoader.prototype.loadExpressionMatrix_ = function(matrixName,
   };
 
   $.get(genotet.data.serverURL, params, function(data) {
-    // Store the last applied data selectors.
-    _(data).extend({
-      matrixname: matrixName,
-      geneRegex: geneRegex,
-      conditionRegex: conditionRegex
-    });
+      // Store the last applied data selectors.
```
yubowenok: Look like extra spaces?

components/expression/expression-loader.js
```diff
@@ -28,8 +30,9 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
  * @param {string} conditionRegex Regex for experiment condition selection.
  * @override
  */
-genotet.ExpressionLoader.prototype.load = function(matrixName, geneRegex,
-    conditionRegex) {
+genotet.ExpressionLoader.prototype.load = function(matrixName, geneRegex, conditionRegex) {
+  this.data.genes = geneRegex.split('|');
```
yubowenok: geneRegex and conditionRegex can be any regex. It doesn't make sense to get genes by splitting. You will get list of gene names and condition names from ajax returns. See [wiki](https://github.com/ViDA-NYU/genotet/wiki/Server-Queries-Documentation).

components/expression/expression-loader.js
```diff
@@ -52,117 +54,237 @@ genotet.ExpressionLoader.prototype.loadExpressionMatrix_ = function(matrixName,
...
+ */
+genotet.ExpressionLoader.prototype.updateGenes = function(method, geneRegex, originParams) {
+  var regex = this.data.matrix.geneRegex;
+  geneRegex = geneRegex.toUpperCase();
+  switch (method) {
+    case 'setGene':
+      // Totally replace the regex.
+      regex = geneRegex;
+      this.data.genes = [];
+      this.data.genes.push(geneRegex);
```
yubowenok: What exactly is genes?  Is it an array of gene names? But you are pushing regex into the array and this makes me confusing. What happenes if geneRegex = aaa|bbb|ccc or even ^[a-z]+[0-9]\*  ?

components/expression/expression-loader.js
```diff
@@ -52,117 +54,237 @@ genotet.ExpressionLoader.prototype.loadExpressionMatrix_ = function(matrixName,
...
+    case 'removeGene':
+      // Remove the regex.
+      var index = this.data.genes.indexOf(geneRegex);
+      if (index == -1) {
+        genotet.error('invalid gene regex', geneRegex);
+        return;
+      };
+      this.data.genes.splice(index, 1);
+      regex = '';
+      this.data.genes.forEach(function(geneRegexs, i) {
```
yubowenok: You may replace "forEach" by

``` js
this.data.genes.join('|');
```

components/expression/expression-loader.js
```diff
@@ -52,117 +54,237 @@ genotet.ExpressionLoader.prototype.loadExpressionMatrix_ = function(matrixName,
...
+        regex += geneRegexs + (i == this.data.genes.length - 1 ? '' : '|');
+      }, this);
+      break;
   }
+  var conditionRegex = '';
```
yubowenok: ditto, and for all below

components/expression/expression-loader.js
```diff
@@ -52,117 +54,237 @@ genotet.ExpressionLoader.prototype.loadExpressionMatrix_ = function(matrixName,
...
+    conditionRegex += conditionNames + (i == this.data.conditions.length - 1 ? '' : '|');
+  }, this);
+  this.signal('loadStart');
+  var params = {
+    type: 'expression',
+    mat: originParams.matrixName,
+    exprows: regex,
+    expcols: conditionRegex
+  };
+  $.get(genotet.data.serverURL, params, function(data) {
```
yubowenok: This seems to be exactly the same ajax routine as "load()". If you are creating regex's that represent the new row/column subsets, please re-use load().

components/expression/expression-loader.js
```diff
@@ -52,117 +54,237 @@ genotet.ExpressionLoader.prototype.loadExpressionMatrix_ = function(matrixName,
...
-  this.parentView.viewdata.lineData = new Array();
-  this.lastIdentifier = identifier;
-  if (identifier.name != null && identifier.name != '') {
-    this.loadLine(identifier.mat, identifier.name);
+/**
+ * Updates the conditions in the current expression.
+ * @param {string} method Update method, either 'set' or 'add'.
+ * @param {string} conditionRegex Regex that selects the conditions to be updated.
+ */
+genotet.ExpressionLoader.prototype.updateConditions = function(method, conditionRegex, originParams) {
```
yubowenok: Are there significant differences between updating genes and conditions? If not, you probably want to combine the two update functions into one.

components/expression/expression-loader.js
```diff
@@ -52,117 +54,237 @@ genotet.ExpressionLoader.prototype.loadExpressionMatrix_ = function(matrixName,
...
+        geneRegex: geneRegex,
+        conditionRegex: regex
+      });
+      this.data.matrix = data;
+      this.signal('loadComplete');
+    }.bind(this), 'jsonp')
+    .fail(this.fail.bind(this, 'cannot update conditions in the expression', params));
 };
 
-LoaderHeatmap.prototype.loadComplete = function() {
```
yubowenok, line 95: Please remove all the old codes below. It doens't seem that they will be re-used. Thanks.

components/expression/expression-panel.html
```diff
@@ -31,7 +43,7 @@
       </tr>
       <tr>
         <td>Condition Labels</td>
-        <td><input type="checkbox" id="label-conditions" title="show experiment condition (condition) labels" checked></td>
```
yubowenok, line 34: I think it makes more sense to check both genes and condition labels by default.

components/expression/expression-panel.js
```diff
@@ -68,7 +71,54 @@ genotet.ExpressionPanel.prototype.initPanel = function() {
...
+      this.signal('update', {
+        type: 'gene',
+        regex: geneRegex,
+        method: method
+      });
+    }.bind(this));
+  }, this);
+
+  // Condition update
+  ['setCondition', 'addCondition', 'removeCondition'].forEach(function(method) {
```
yubowenok, line 106: You may combine this "condition update" block with the "gene update" block above. Most of the codes are duplicated.

components/expression/expression-panel.js
```diff
@@ -93,19 +143,109 @@ genotet.ExpressionPanel.prototype.updateGenes = function(genes) {
...
+ * @private
+ */
+genotet.ExpressionPanel.prototype.hideCellInfo_ = function() {
+  this.container_.find('#cell-info').slideUp();
+};
+
+/**
+ * Displays a tooltip around cursor about a hovered cell.
+ * @param {!String} geneName Gene Name being hovered.
+ * @param {!String} conditionName Condition Name being hovered.
```
yubowenok, line 176: Please mark the private functions "@private".

components/expression/expression-panel.html
```diff
@@ -48,4 +60,26 @@
...
+    <div id="gene">
+      <label>Gene</label>
+      <span></span>
+    </div>
+    <div id="condition">
+      <label>Condition</label>
+      <span></span>
+    </div>
+  </div>
+  <div id="path-info" class="panel-section box" style="display: none;">
```
yubowenok: What is "path-info"? If this is for profiles, then probably you should name it "profile-info".

components/expression/expression-renderer.js
```diff
@@ -172,20 +251,85 @@ genotet.ExpressionRenderer.prototype.drawExpressionMatrix_ = function() {
...
+    transformLeft += this.LABEL_MARGIN_;
+  }
+  else if (this.data.options.showProfiles) {
+    transformLeft += this.DEFAULT_PROFILE_MARGIN;
+  }
+  if (this.data.options.showConditionLabels) {
+    transformTop += this.LABEL_MARGIN_;
+  }
+
+  this.svgHeatmapContent_.selectAll('g').remove();
```
yubowenok: Please avoid clearing screen.

components/expression/expression-renderer.js
```diff
@@ -198,18 +342,29 @@ genotet.ExpressionRenderer.prototype.drawMatrixGeneLabels_ = function() {
...
+    .attr("transform", genotet.utils.getTransform([
+      this.geneLabelWidth_,
+      transformTop
+    ]))
+    .text(_.identity)
+    .attr('x', 0)
+    .attr('y', function(geneName, i) {
+      return i * cellHeight;
+    })
+    .style('text-anchor', 'end');
```
yubowenok: Please style this in the css, as it applies for all such labels.

components/expression/expression-renderer.js
```diff
@@ -244,180 +413,341 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+    .attr(
+      'transform',
+      'translate(0, ' + (this.profileHeight_ - this.PROFILE_MARGINS.BOTTOM) + ')'
+    );
+  this.svgProfile_
+    .append('svg:g').call(yAxis)
+    .classed('axis', true)
+    .attr('transform', 'translate(' + this.PROFILE_MARGINS.LEFT + ', 0)');
+
+  var colors = d3.scale.category20()
```
yubowenok: Maybe create such a "category60" array in data and re-use it.

components/expression/expression-renderer.js
```diff
@@ -244,180 +413,341 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+    .x(function(data, i) {
+      return xScale(i);
+    })
+    .y(function(data) {
+      return yScale(data);
+    })
+    .interpolate('linear');
+  this.data.profiles.forEach(function(path, i) {
+    var geneIndex = this.data.profiles[i].row;
+    var pathColor = colors[genotet.utils.hashString(heatmapData.geneNames[geneIndex]) % 60];
```
yubowenok: Please avoid constant "60".

components/expression/expression-renderer.js
```diff
@@ -244,180 +413,341 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
   }
+  else if (!this.data.options.showGeneLabels) {
+    this.PROFILE_MARGINS.LEFT = this.DEFAULT_PROFILE_MARGIN;
+  }
+  else {
+    this.PROFILE_MARGINS.LEFT = this.geneLabelWidth_ + this.LABEL_MARGIN_;
+  }
+
+  var heatmapData = this.data.matrix;
+  this.svgProfile_.selectAll('g').remove();
```
yubowenok, line 466: Please avoid clearing screens. Always use enter(), exit() and update routines for data elements (that could potentially be many).

components/expression/expression-renderer.js
```diff
@@ -244,180 +413,341 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+    .attr(
+      'height',
+      this.profileHeight_ - this.PROFILE_MARGINS.TOP - this.PROFILE_MARGINS.BOTTOM
+    );
+
+  var line = d3.svg.line()
+    .x(function(data, i) {
+      return xScale(i);
+    })
+    .y(function(data) {
```
yubowenok: .y(yScale)

components/expression/expression-renderer.js
```diff
@@ -172,20 +251,85 @@ genotet.ExpressionRenderer.prototype.drawExpressionMatrix_ = function() {
...
+    .attr('height', cellHeight)
+    .attr('x', function(value, i) {
+      return i * cellWidth;
+    })
+    .attr('y', function(velue, i, j) {
+      return j * cellHeight;
     })
-    .attr('height', function(cell) {
-      return cell.height;
+    .style('stroke', function(value) {
```
yubowenok: .style('stroke', colorScale)

components/expression/expression-renderer.js
```diff
@@ -244,180 +413,341 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+      geneName: heatmapData.geneNames[geneIndex],
+      row: geneIndex,
+      color: pathColor
+    };
+    profileContent.append('path')
+      .attr('d', line(heatmapData.values[geneIndex]))
+      .attr('stroke', pathColor)
+      .attr('fill', 'none')
+      .on('mouseover', function(value) {
+        this.data.profiles[i].container_= d3.event.target;
```
yubowenok, line 530: It looks weird to set container for event callback. An element's container is usually set when the element is created and never changed.

components/expression/expression-renderer.js
```diff
@@ -244,180 +413,341 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+        this.signal('pathHover', this.data.profiles[i]);
+      }.bind(this))
+      .on('mouseout', function(value) {
+        this.data.profiles[i].container_= d3.event.target;
+        this.signal('pathUnhover', this.data.profiles[i]);
```
yubowenok, line 538: Please replace "path" by "profile". "path" is an implementation specific word. However data specifc words are preferred for readability.

components/expression/expression-renderer.js
```diff
@@ -244,180 +413,341 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
-genotet.ExpressionRenderer.prototype.resize = function() {
-  this.base.resize.call(this);
-  this.render();
+/**
+ * Adds the expression profiles for the selected genes as line charts.
+ * @private
+ */
+genotet.ExpressionRenderer.prototype.removeGeneProfile_ = function(geneIndex) {
+  var index = -1;
+  this.data.profiles.forEach(function(path, i) {
```
yubowenok: Use a for loop and break when you find an element may be better in this case.

Or if you don't mind being a bit not efficient, but much cleaner, do

``` js
var index = this.data.profiles.map(function(path) {
  return path.row
}).indexOf(geneIndex);
```

components/expression/expression-renderer.js
```diff
@@ -244,180 +413,341 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
-    }
-    this.lines[i].color = this.lineColors[i];
+/**
+ * Highlights the hover cell for the heatmap.
+ * @private
+ */
+genotet.ExpressionRenderer.prototype.highlightHoverCell_ = function(cell, highlight_) {
+  var cellSelection = d3.select(cell.container_);
+  if (highlight_) {
+    cellSelection.style('stroke', 'white');
```
yubowenok: Re-classing every cell's 'text-highlight' to highlight a single cell doesn't seem to be a good idea...

components/expression/expression-renderer.js
```diff
@@ -244,180 +413,341 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+    this.svgGeneLabels_.selectAll('text').classed('text-highlight', function(d, i) {
```
yubowenok: You can name the class "highlighted". In your css you can do

``` css
text.highlighted {
  stroke: ...
}
```

components/expression/expression-renderer.js
```diff
@@ -244,180 +413,341 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
-    this.svg.selectAll('#tfacondtext').style('text-anchor', 'start');
+/**
+ * Highlights the hover path for the gene profile.
+ * @private
+ */
+genotet.ExpressionRenderer.prototype.highlightHoverPath_ = function(path, highlight_) {
+  var pathSelection = d3.select(path.container_);
+  if (highlight_) {
+    pathSelection
+      .style('stroke', 'orange')
```
yubowenok: Please move these constant settings to css.

components/expression/expression-renderer.js
```diff
@@ -244,180 +413,341 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
-  this.parentView.loader.loadHeatmap(null, exprows, expcols);
-  $('#'+ this.htmlid + ' #layoutwrapper #selregion').remove();
+/** @inheritDoc */
+genotet.ExpressionRenderer.prototype.resize = function() {
+  this.base.resize.call(this);
+  this.render(this.data.options.autoScaleGradient);
 };
-*/
+
+/*
```
yubowenok, line 692: Please remove the old code. Thanks.

components/expression/expression-renderer.js
```diff
@@ -116,27 +186,36 @@ genotet.ExpressionRenderer.prototype.layout = function() {
...
       this.geneLabelWidth_,
-      this.heatmapHeight_
+      0
     ]));
 };
 
 /** @inheritDoc */
 genotet.ExpressionRenderer.prototype.dataLoaded = function() {
-  this.render();
+  this.render(this.data.options.autoScaleGradient);
```
yubowenok: render() is a general function. It shouldn't be accepting parameters like autoScaleGradient, which it could get directly from this.data.options.

components/expression/expression.css
```diff
@@ -1,3 +1,23 @@
+.text-highlight {
```
yubowenok: See above. Use sth like "text.highlighted".

components/expression/expression.js
```diff
@@ -30,6 +30,70 @@ genotet.ExpressionView = function(viewName, params) {
...
+      this.renderer.addGeneProfile_(geneIndex);
+    }.bind(this))
+    .on('genotet.removeGeneProfile', function(event, geneIndex) {
+      this.renderer.removeGeneProfile_(geneIndex);
+    }.bind(this));
+
+  // Cell hover in expression.
+  $(this.renderer)
+    .on('genotet.cellHover', function(event, cell) {
+      this.renderer.highlightHoverCell_(cell, true);
```
yubowenok: It is recommended to sperate highlight and unhighlight into two functions, rather than passing a boolean parameter.

js/panel-manager.js
yubowenok: Pending fixes. Is there a place where you set this to "true"? If not you probably should remove this.

js/utils.js
```diff
@@ -208,13 +208,51 @@ genotet.utils.multiplyVector = function(p, k) {
...
+  }
+  return result;
+};
+
+/**
+ * Hashes a string to a color.
+ * @param {string} s
+ * @return {number} Hash value is a 6-digit color code.
+ */
+genotet.utils.hashStringToColorCode = function(s) {
```
yubowenok: This doesn't seem to be used? If it is not used please remove. 

Where is the hash algorithm from?

components/expression/expression-loader.js
liana-yang: Of course they are! They are used for gene and condition update separately in the panel manager.

components/expression/expression-panel.html
liana-yang, line 34: It would lead to errors if we check the condition labels by default, because the view window is not large enough to display the whole labels.

components/expression/expression-panel.js
liana-yang, line 106: ditto.

components/expression/expression-panel.html
liana-yang: It should be profile-info... My mistake.

components/expression/expression-renderer.js
liana-yang: I think for loop is better.

components/expression/expression-panel.html
yubowenok, line 34: It does not mean that the labels should be unchecked by default. You may want to resolve the height error using other method.

components/expression/expression-loader.js
yubowenok: I know you used your two functions (of course) but this is not SIGNIFICANT difference. Your update gene/condition routines are extremely similar (aren't they?) and IMO they should be combined into one to prevent duplicated code.

components/expression/expression-loader.js
liana-yang: call load()

components/expression/expression-loader.js
Shauro: Fixed.

Shauro: Fixed.

components/expression/expression-loader.js
Shauro: Fixed.

components/expression/expression-loader.js
liana-yang: Fixed.

components/expression/expression-loader.js
liana-yang: Fixed.

components/expression/expression-loader.js
liana-yang: Fixed.

components/expression/expression-loader.js
liana-yang: Fixed.

components/expression/expression-loader.js
liana-yang, line 95: Fixed.

components/expression/expression-panel.html
liana-yang: Fixed.

components/expression/expression-panel.js
Shauro: Fixed.

components/expression/expression-panel.js
liana-yang, line 176: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
liana-yang, line 466: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
liana-yang, line 530: The event container is not a container. It is an path element, which has a container attribute.

liana-yang, line 530: Fixed.

components/expression/expression-renderer.js
liana-yang, line 538: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression-renderer.js
liana-yang: Fixed.

components/expression/expression.css
liana-yang: Fixed.

components/expression/expression.js
liana-yang: Fixed.

js/utils.js
liana-yang: Fixed.

components/expression/expression-panel.js
liana-yang, line 106: Actually I don't think merge them is a good way. Because they are much more different at selection names. Merge would cost more codes to check whether to revise gene or condition.

PR #31, opened 2015-12-07
add listing on creating new view

djasmine:
some bug fix and add listing function on creating new views.


yubowenok:
This doesn't seem to work. The data listings upon node creation are not shown properly. PLEASE DO NOT PULL REQUEST UNTESTED AND NON-WORKING CODE.

PR #39, opened 2015-12-16
Added expression legend and gradient. Added panel tab auto resize.

liana-yang:


server/server.js
```diff
@@ -59,7 +59,7 @@ var bedPath;
  * Reads the configuration file and gets the file paths.
  */
 function config() {
-  var tokens = fs.readFileSync('server/config')
+  var tokens = fs.readFileSync('config')
```
yubowenok: Please remove this change. The server is currently started as server/server.js in the genomes root folder. Or you can change the readme to reflect your change here.

Jiaming changed this to server/server.js originally. Im not sure which way is better.

src/components/expression/expression-renderer.js
```diff
@@ -87,6 +88,12 @@ genotet.ExpressionRenderer = function(container, data) {
   this.LABEL_MARGIN_ = 10;
 
   /**
+   * The difference between the labels and the heatmap.
+   * @private {number}
+   */
+  this.LABEL_DIFFERENCE_ = 10;
```
yubowenok: These constants are used throughout the same class, and may be defined under prototype.

src/components/expression/expression-renderer.js
```diff
@@ -243,6 +281,8 @@ genotet.ExpressionRenderer.prototype.layout = function() {
       this.geneLabelWidth_,
       0
     ]));
+  this.svgHeatmapGradient_
+    .attr('transform', genotet.utils.getTransform([0, 0]));
```
yubowenok: Seems redundant to transform (0, 0).

src/components/expression/expression-renderer.js
```diff
@@ -318,7 +359,28 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+
+  if (this.data.clickedCell) {
+    this.data.clickedCell.row = heatmapData.geneNames.indexOf(
+      this.data.clickedCell.geneName
+    );
+    this.data.clickedCell.column = heatmapData.conditionNames.indexOf(
+      this.data.clickedCell.conditionName
+    );
+    if (
+      this.data.clickedCell.row == -1 || this.data.clickedCell.column == -1
```
yubowenok: The first condition should be on the first line, and the second on the second line. The OR operator is at the end of the first line.

src/components/expression/expression-renderer.js
```diff
@@ -318,7 +359,28 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+    ) {
+      this.unhighlightLabelsForClickedCell_();
+      this.signal('cellUnclick');
+    }
+    else if (
```
yubowenok: Else if needs to be on the previous line.

src/components/expression/expression-renderer.js
```diff
@@ -389,10 +451,11 @@ genotet.ExpressionRenderer.prototype.drawMatrixGeneLabels_ = function() {
   var heatmapData = this.data.matrix;
   var geneLabelsData = heatmapData.geneNames;
   var cellHeight = this.heatmapHeight_ / geneLabelsData.length;
-  var transformTop = this.profileHeight_ + this.TEXT_HEIGHT_ / 2 +
+  var transformTop = this.profileHeight_ + this.TEXT_HEIGHT_ / 3 +
     cellHeight / 2;
```
yubowenok: Why dividing over 3 instead of 2. Please explain.

src/components/expression/expression-renderer.js
```diff
@@ -452,6 +515,99 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
+      genotet.utils.getTransform([transformLeft, transformTop])
+    );
+  var gradientHeight = this.DEFAULT_HEATMAP_GRADIENT_HEIGHT -
+    this.HEATMAP_GRADIENT_MARGINS_.TOP -
+    this.HEATMAP_GRADIENT_MARGINS_.BOTTOM;
+  var gradientWidth = this.heatmapWidth_ - marginLeft - marginRight;
+  var gradientContent = this.svgHeatmapGradient_.append('defs')
+    .append('linearGradient')
+    .attr('id', 'gradient-content')
+    .attr('x1', '0%')
```
yubowenok: Is 0% redundant?

src/components/expression/expression-renderer.js
```diff
@@ -474,13 +630,41 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
   this.svgProfile_.selectAll('g').remove();
   this.svgProfile_.attr('width', this.canvasWidth_);
 
+  var legend = this.svgProfile_.append('g')
+    .classed('legend', true)
+    .attr('transform', genotet.utils.getTransform([
+      this.PROFILE_MARGINS_.LEFT,
+      this.DEFAULT_PROFILE_LEGEND_MARGIN]));
```
yubowenok: The square bracket needs to be on a new line.

src/components/expression/expression-renderer.js
```diff
@@ -452,6 +515,99 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
+  else if (this.data.options.showProfiles) {
+    transformLeft += this.DEFAULT_PROFILE_MARGIN;
+  }
+  if (this.data.options.showConditionLabels) {
+    transformTop += this.conditionLabelHeight_ +
+      this.LABEL_MARGIN_ + this.LABEL_DIFFERENCE_;
+  }
+  this.svgHeatmapGradient_
+    .attr(
+      'transform',
```
yubowenok: This needs to be on a previous line.

src/components/expression/expression-renderer.js
```diff
@@ -452,6 +515,99 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
+genotet.ExpressionRenderer.prototype.drawHeatmapGradient_ = function() {
+  this.svgHeatmapGradient_.selectAll('*').remove();
+  var heatmapData = this.data.matrix;
+  var scaleMin = 0;
+  var scaleMax = 0;
+  if (this.data.options.autoScaleGradient) {
+    scaleMin = heatmapData.valueMin;
+    scaleMax = heatmapData.valueMax;
+  }
+  else {
```
yubowenok: Else needs to be on the previous line

src/components/expression/expression-renderer.js
```diff
@@ -452,6 +515,99 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
+  }
+  if (this.data.options.showConditionLabels) {
+    transformTop += this.conditionLabelHeight_ +
+      this.LABEL_MARGIN_ + this.LABEL_DIFFERENCE_;
+  }
+  this.svgHeatmapGradient_
+    .attr(
+      'transform',
+      genotet.utils.getTransform([transformLeft, transformTop])
+    );
```
yubowenok: and the closing bracket following: ...getTransform(...));

``` js
.attr('transform',
  genotet.utils.getTransform([transformLeft, transformTop]));
```

src/components/expression/expression-renderer.js
```diff
@@ -452,6 +515,99 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
+  var gradientHeight = this.DEFAULT_HEATMAP_GRADIENT_HEIGHT -
+    this.HEATMAP_GRADIENT_MARGINS_.TOP -
+    this.HEATMAP_GRADIENT_MARGINS_.BOTTOM;
+  var gradientWidth = this.heatmapWidth_ - marginLeft - marginRight;
+  var gradientContent = this.svgHeatmapGradient_.append('defs')
```
yubowenok: I'd recommend writing this gradient as a template somewhere in the node's template. It's your call though.

src/components/expression/expression-renderer.js
```diff
@@ -452,6 +515,99 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
+      this.HEATMAP_GRADIENT_MARGINS_.TOP
+      ]))
+    .attr('width', gradientWidth)
+    .attr('height', gradientHeight)
+    .style('fill', 'url(#gradient-content)');
+  this.svgHeatmapGradient_.append('text')
+    .attr('x', marginLeft - 1)
+    .attr('y',
+      this.HEATMAP_GRADIENT_MARGINS_.TOP +
+      gradientHeight / 2 + this.TEXT_HEIGHT_ / 3)
```
yubowenok: It's bad practice to have magical number "over 3" everywhere.

src/components/expression/expression-renderer.js
```diff
@@ -474,13 +630,41 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
   this.svgProfile_.selectAll('g').remove();
```
yubowenok: You seem to be redrawing again?

src/components/expression/expression-renderer.js
```diff
@@ -474,13 +630,41 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+    }
+  }
+
   var xScale = d3.scale.linear().range([
     this.PROFILE_MARGINS_.LEFT,
     this.canvasWidth_ - this.PROFILE_MARGINS_.RIGHT
   ]).domain([0, heatmapData.conditionNames.length]);
+  var yScaleTop = this.PROFILE_MARGINS_.TOP +
+    this.DEFAULT_PROFILE_LEGEND_HEIGHT;
+  if (this.data.profiles.length == 0) {
```
yubowenok, line 727: I think this should be computed in layout(), not here.

src/components/expression/expression-renderer.js
```diff
@@ -490,15 +674,17 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
     .append('svg:g').call(xAxis)
     .classed('axis', true)
```
yubowenok: You can do:

``` js
.classed('x axis', true);
```

src/components/expression/expression-renderer.js
```diff
@@ -452,6 +515,99 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
+    .attr('offset', '50%')
+    .attr('stop-color', genotet.data.redYellowScale[1])
+    .attr('stop-opacity', 1);
+  gradientContent.append('stop')
+    .attr('offset', '100%')
+    .attr('stop-color', genotet.data.redYellowScale[2])
+    .attr('stop-opacity', 1);
+  this.svgHeatmapGradient_.append('rect')
+    .classed('gradient-rect', true)
+    .attr('transform', genotet.utils.getTransform([
```
yubowenok, line 682:
``` js
.attr('transform', genotet.utils.getTransform([
  marginLeft,
  this.HEATMAP_GRADIENT_MARGINS_.TOP
]))
```

src/components/expression/expression-renderer.js
```diff
@@ -518,34 +704,45 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+      profile.geneName.length * this.GENE_LABEL_WIDTH_FACTOR_);
+    legend.append('rect')
+      .attr('height', legendHeight)
+      .attr('width', legendHeight)
+      .attr('x', legendWidth)
+      .style('fill', pathColor);
+    legend.append('text')
+      .text(profile.geneName)
+      .attr('transform', genotet.utils.getTransform([
+        legendHeight + 1,
```
yubowenok: Please explain the "+1" by either comments or using a self-explained constant name.

src/components/expression/expression.scss
```diff
@@ -44,3 +44,8 @@
     }
   }
 }
+
+.gradient-rect {
+  rx: 10;
```
yubowenok: rects don't seem to have rx, ry. Besides attr properties in css won't work I think.

src/panel-manager.js
```diff
@@ -9,6 +9,14 @@ genotet.panelManager = {};
 genotet.panelManager.COLLAPSED_WIDTH = 32;
 /** @const {number} */
 genotet.panelManager.TRANSITION_TIME = 300;
+/** @const {number} */
+genotet.panelManager.TOGGLE_BTN_HEIGHT = 50;
+/** @const {number} */
+genotet.panelManager.TAB_MARGIN_BOTTOM_DIFFERENCE = 20;
```
yubowenok: Please explain the word "difference".

src/panel-manager.js
```diff
@@ -9,6 +9,14 @@ genotet.panelManager = {};
 genotet.panelManager.COLLAPSED_WIDTH = 32;
 /** @const {number} */
 genotet.panelManager.TRANSITION_TIME = 300;
+/** @const {number} */
```
yubowenok: All these constants look @private.

src/panel-manager.js
```diff
@@ -102,8 +111,12 @@ genotet.panelManager.addPanel = function(view) {
...
   $(view).on('genotet.focus', function() {
     var clickedViewID = view.viewID_;
-    genotet.panelManager.activatePanel_(clickedViewID);
-  });
+    this.activatePanel_(clickedViewID);
+  }.bind(this));
+  this.adjustTabHeight();
+
+  // Adjust tab height.
+  $(window).resize(this.adjustTabHeight.bind(this));
```
yubowenok: This doesn't seem right. You are binding window resize handler EVERY TIME a view is created. It should be binded only once.

src/panel-manager.js
```diff
@@ -86,6 +94,7 @@ genotet.panelManager.activatePanel_ = function(viewID) {
  * Creates a panel with the given name.
  * @param {!genotet.View} view Name of the view.
  * @return {!jQuery}
+ * @this
```
yubowenok: You need to specify what @this is. Otherwise it is meaningless. You may do @this {genotet.panelManager} here. However, we need to avoid using namespace as @this. You may want to replace all "this" by "genotet.panelManager".

src/panel-manager.js
```diff
@@ -134,6 +147,7 @@ genotet.panelManager.addPanel = function(view) {
 /**
  * Closes the given panel.
  * @param {string} viewName Name of the view.
+ * @this
```
yubowenok: ditto. @this without typing is useless but only cheating the compiler... Please fix thoughout this file.

src/panel-manager.js
```diff
@@ -142,6 +156,7 @@ genotet.panelManager.removePanel = function(viewName) {
   var activated = tab.hasClass('active');
   tab.remove();
   panel.remove();
+  this.adjustTabHeight();
```
yubowenok: use "genotet.panelManager" instead

src/panel-manager.js
```diff
@@ -171,3 +186,30 @@ genotet.panelManager.closeAllPanels = function() {
...
+ * Adjust tab height.
+ * @this
+ */
+genotet.panelManager.adjustTabHeight = function() {
+  var tabSelector = $('.sideways');
+  var tabCount = tabSelector.find('li').length - 1;
+  var tabHeight = tabSelector.find('li:nth-child(2) a').outerWidth();
+  var tabContentHeight = $('.tab-content').outerHeight();
+  var newTabHeight = Math.floor((tabContentHeight -
+    this.TAB_HEIGHT_DIFFERENCE - this.TOGGLE_BTN_HEIGHT) / tabCount - 1
```
yubowenok: Please explain "-1".

server/server.js
liana-yang: Fixed.

src/components/expression/expression-renderer.js
liana-yang: Fixed.

src/components/expression/expression-renderer.js
liana-yang: Fixed.

src/components/expression/expression-renderer.js
liana-yang: Fixed.

src/components/expression/expression-renderer.js
liana-yang: Fixed.

src/components/expression/expression-renderer.js
liana-yang: Fixed.

src/components/expression/expression-renderer.js
liana-yang: Fixed.

src/components/expression/expression-renderer.js
liana-yang: Fixed.

src/components/expression/expression-renderer.js
Shauro: Fixed.

src/components/expression/expression-renderer.js
Shauro: Fixed.

src/components/expression/expression-renderer.js
liana-yang: Fixed.

src/components/expression/expression-renderer.js
liana-yang, line 727: Fixed.

src/components/expression/expression-renderer.js
liana-yang: Fixed.

src/components/expression/expression.scss
liana-yang: It works. http://www.w3schools.com/svg/svg_rect.asp

yubowenok: Okay I don't know rects have rx,ry :)  Then please set it via attr to avoid browser compatibilty issue, e.g. rx in style doesn't seem to work in IE11.

src/panel-manager.js
liana-yang: It represents the different value between the width and margin-bottom of the tabs container. Because the margin-bottom value has to be adjusted with the width value.

src/panel-manager.js
liana-yang: Why @private? They are constants and we should use @const to indicate a constant (non-overwritable) pointer (a variable or property) according to JsDoc.

src/panel-manager.js
Shauro: I am sorry that I don't understand why we need to avoid using namespace as @this. Could you please explain it in detail? Thank you very much!

liana-yang: Why we need to avoid using namespace as @this?

src/panel-manager.js
yubowenok: I mean explain in comments not here...

src/panel-manager.js
yubowenok:
``` js
genotet.panelManager.TRANSITION_TIME_ = 300;
```

Because it seems that these variables won't be used outside this file.

src/panel-manager.js
yubowenok: @this usually points to a class instance. genotet.panelManager is a namespace, not a class instance.

src/panel-manager.js
liana-yang: Because there exists an initial invisible panel which is used to be cloned when adding new panels. We should exclude it when calculate new tab height.

yubowenok: Please explain in the comments. Think of a future reader of this code who is going to either use or modify this code, how can he/she understand the magic numbers? :)

yubowenok:
I recommend syncing with refactor first and then go on making more changes. Always sync with base when you are behind so that you can avoid modifying out-of-date stuffs.

src/components/expression/expression-renderer.js
liana-yang, line 682: Fixed.

yubowenok, line 682: BTW if the gradient is always in a format like above, can't we make a static svg DOM and set its attr and visibility on demand? In this way we can avoid appending the same stuff repeatedly and reduce "hard coded html" in js source. Up to you though.

liana-yang, line 682: Fixed. :)

src/components/expression/expression-renderer.js
```diff
@@ -74,73 +73,45 @@ genotet.ExpressionRenderer = function(container, data) {
...
-  this.LABEL_MARGIN_ = 10;
+  this.profileHeight_ = this.DEFAULT_PROFILE_HEIGHT;
 
   /**
-   * The height of the label text.
+   * Width of the heatmap.
    * @private {number}
    */
-  this.TEXT_HEIGHT_ = 14.5;
+    this.heatmapWidth_ = 0;
```
yubowenok: extra leading spaces?

src/components/expression/expression-renderer.js
```diff
@@ -74,73 +73,45 @@ genotet.ExpressionRenderer = function(container, data) {
...
-    .concat(d3.scale.category20c().range());
+  this.heatmapHeight_ = 0;
 
   /**
-   * Margins of the gene profile plot.
+   * Margins of the gene profile.
    * @private {!Object}
    */
-  this.PROFILE_MARGINS_ = {
+  this.profileMargins_ = {
```
yubowenok, line 66: Umm if this is not constant then the attribute names should be lowercased.

src/components/expression/expression-renderer.js
```diff
@@ -74,73 +73,45 @@ genotet.ExpressionRenderer = function(container, data) {
...
-  this.heatmapHeight_ = 0;
-
-  /**
-   * Margin of the gene profile.
-   * @private {number}
+   * Margins of the heatmap gradient.
+   * @private {!Object}
    */
-  this.geneProfileMargin_ = 0;
+  this.heatmapGradientMargins_ = {
```
yubowenok: ditto

src/components/expression/expression-renderer.js
```diff
@@ -149,24 +120,46 @@ genotet.utils.inherit(genotet.ExpressionRenderer, genotet.ViewRenderer);
...
- this.heatmapFontHSize = 8.5;
- */
+/** @const {number} */
+genotet.ExpressionRenderer.prototype.DEFAULT_HEATMAP_GRADIENT_HEIGHT = 30;
+
+/** @const {number} */
+genotet.ExpressionRenderer.prototype.GENE_LABEL_WIDTH_FACTOR = 8.725;
 
+/** @const {number} */
+genotet.ExpressionRenderer.prototype.CONDITION_LABEL_HEIGHT_FACTOR = 6.501175;
```
yubowenok, line 197: This number is too magical :) I guess the best way is to pre-render (or say fake render) the labels and determine the exact on-screen pixel width/height of the font. 

Anyway it is not very important so we can leave it for now.

src/components/expression/expression-renderer.js
```diff
@@ -210,6 +203,51 @@ genotet.ExpressionRenderer.prototype.initLayout = function() {
...
+   * @private {!d3.selection}
+   */
+  this.svgHeatmapGradient_ = this.svgHeatmap_.append('g')
+    .classed('heatmap-gradient', true);
+
+  /**
+   * SVG defs for the heatmap gradient.
+   * @private {!d3.selection}
+   */
+  this.gradientContent_ = this.svgHeatmapGradient_.append('defs')
```
yubowenok, line 305: Why don't we write sth like

``` html
<defs>
  <linearGradient id="gradient-content" x1="0%" x2="100%" spreadMethod="pad" ...></linearGradient>
</defs>
```

in the html svg and avoid "appending" in js?

src/components/expression/expression-renderer.js
```diff
@@ -310,15 +350,35 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+  if (this.data.clickedCell) {
+    this.data.clickedCell.row = heatmapData.geneNames.indexOf(
+      this.data.clickedCell.geneName
+    );
+    this.data.clickedCell.column = heatmapData.conditionNames.indexOf(
+      this.data.clickedCell.conditionName
+    );
+    if (this.data.clickedCell.row == -1 ||
+      this.data.clickedCell.column == -1
+    ) {
```
yubowenok: It seesm that this doesn't need to start a new line?

``` js
    if (this.data.clickedCell.row == -1 || this.data.clickedCell.column == -1) {
       // ...
    }
```

exactly 80 chars? :)

src/components/expression/expression-renderer.js
```diff
@@ -310,15 +350,35 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+      this.unhighlightLabelsForClickedCell_();
+      this.signal('cellUnclick');
+    } else if (
+      this.data.clickedCell.row != -1 && this.data.clickedCell.column != -1
```
yubowenok: not sth critical but pls try to end lines with operators like "&&" and "+", and not brackets "("

``` js
    } else if (this.data.clickedCell.row != -1 &&
      this.data.clickedCell.column != -1)
```

src/panel-manager.js
```diff
@@ -37,6 +50,7 @@ genotet.panelManager.init = function() {
       genotet.panelManager.togglePanel_();
     }
   });
+  $(window).resize(this.adjustTabHeight.bind(this));
```
yubowenok:
``` js
$(window).resize(genotet.panelManager.adjustTabHeight);
```

and avoid this throughout this file.

src/panel-manager.js
```diff
@@ -169,3 +188,35 @@ genotet.panelManager.closeAllPanels = function() {
 
   $('#side-panel').hide();
 };
+
+/**
+ * Adjust tab height.
+ * @this {genotet.panelManager}
```
yubowenok: ditto

src/panel-manager.js
```diff
@@ -88,6 +102,7 @@ genotet.panelManager.activatePanel_ = function(viewID) {
  * Creates a panel with the given name.
  * @param {!genotet.View} view Name of the view.
  * @return {!jQuery}
+ * @this {genotet.panelManager}
```
yubowenok: remove @this and replace all "this" in this file by genotet.panelManager.

src/components/expression/expression-renderer.js
```diff
@@ -673,23 +853,68 @@ genotet.ExpressionRenderer.prototype.highlightLabelsForClickedCell_ = function(
...
+      .selectAll('text')
+      .classed('click-highlighted', false);
+    this.svgConditionLabels_
+      .selectAll('text')
+      .classed('click-highlighted', false);
+  };
+
+/**
  * Highlights the labels of clicked profile.
  * @param {*} profile TODO(liana): fill the type.
```
yubowenok: TODO?

src/components/expression/expression-renderer.js
```diff
@@ -389,10 +449,11 @@ genotet.ExpressionRenderer.prototype.drawMatrixGeneLabels_ = function() {
   var heatmapData = this.data.matrix;
   var geneLabelsData = heatmapData.geneNames;
   var cellHeight = this.heatmapHeight_ / geneLabelsData.length;
-  var transformTop = this.profileHeight_ + this.TEXT_HEIGHT_ / 2 +
+  var transformTop = this.profileHeight_ + this.TEXT_HEIGHT / 3 +
```
yubowenok: Have you explained "dividing by 3" in comments?

src/components/expression/expression-renderer.js
```diff
@@ -452,6 +513,75 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
+      marginLeft,
+      this.heatmapGradientMargins_.TOP
+    ]))
+    .attr('width', gradientWidth)
+    .attr('height', gradientHeight)
+    .attr('rx', 10)
+    .attr('ry', 10);
+  d3.select('.gradient-text-left')
+    .attr('x', marginLeft - this.HEATMAP_GRADIENT_MARGIN)
+    .attr('y',
```
``` js
.attr('y', this.heatmapGradientMargins_.TOP + gradientHeight / 2 +
   this.TEXT_HEIGHT / 3)
```

Try to put as many operators in a line as possible, unless it is much clearer to start a new line, e.g. for constant array declaration.

src/components/expression/expression-renderer.js
```diff
@@ -452,6 +513,75 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
+      this.heatmapGradientMargins_.TOP +
+      gradientHeight / 2 + this.TEXT_HEIGHT / 3)
+    .text(scaleMin)
+    .style('text-anchor', 'end');
+  d3.select('.gradient-text-right')
+    .attr('x', marginLeft + gradientWidth + this.HEATMAP_GRADIENT_MARGIN)
+    .attr(
+      'y',
```
yubowenok: ditto. this formatting is a bit weird.

src/components/expression/expression-renderer.js
```diff
@@ -462,25 +592,54 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
   }
   else {
```
yubowenok: this else shall be on the previous line

src/components/expression/expression-renderer.js
```diff
@@ -462,25 +592,54 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
     if (!this.data.options.showGeneLabels) {
-      this.PROFILE_MARGINS_.LEFT = this.DEFAULT_PROFILE_MARGIN;
+      this.profileMargins_.LEFT = this.DEFAULT_PROFILE_MARGIN;
     }
     else {
```
yubowenok: this else shall be on the previous line

src/components/expression/expression-renderer.js
```diff
@@ -616,7 +796,7 @@ genotet.ExpressionRenderer.prototype.highlightHoverCell_ = function(cell) {
 
 /**
  * Unhighlights the hover cell for the heatmap.
- * @param {*} cell TODO(liana): fill the type.
+ * @param {!Object} cell
```
yubowenok: Umm this {!Object} type is not very informative. You may want to declare your cell type

``` js
/**
 * @typedef {{
 *   container_: ?jQuery,
 *   geneName: ?string,
 *   conditionName: ?string,
 *   row: number,
 *   column: number,
 *   value: number,
 *   colorscaleValue: ?number
 * }}
 */
genotet.ExpressionRenderer.Cell;
```

https://developers.google.com/closure/compiler/docs/js-for-compiler#tag-typedef

Then you can use this Cell type throughout.

``` js
@param {!genotet.ExpressionRenderer.Cell} cell
```

src/components/expression/expression-renderer.js
```diff
@@ -628,7 +808,7 @@ genotet.ExpressionRenderer.prototype.unhighlightHoverCell_ = function(cell) {
 
 /**
  * Highlights the hover profile for the gene profile.
- * @param {*} profile TODO(liana): fill the type.
+ * @param {!Object} profile
```
yubowenok: Similarly  you may want to typedef genotet.ExpressionRenderer.Profile.

src/components/expression/expression-renderer.js
```diff
@@ -673,23 +853,68 @@ genotet.ExpressionRenderer.prototype.highlightLabelsForClickedCell_ = function(
 };
 
 /**
+ * Unhighlights the label of the clicked cells for the heatmap.
+ * @param {object} cell
```
yubowenok: should be {!Object}, but preferrably {!genotet.ExpressionRenderer.Cell}

src/components/expression/expression-renderer.js
liana-yang, line 197: I calculated it in a magic way that I cannot remember now... = = Okay, let's leave it temporarily.

src/components/expression/expression-renderer.js
```diff
@@ -452,6 +513,75 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
+
+  var transformLeft = 0;
+  var transformTop = this.heatmapHeight_;
+  if (this.data.options.showProfiles) {
+    transformTop += this.profileHeight_;
+  }
+  if (this.data.options.showGeneLabels) {
+    transformLeft += this.geneLabelWidth_ + this.LABEL_MARGIN;
+  }
+  else if (this.data.options.showProfiles) {
```
``` js
if (condition) {
  // something
} else if {
  // something
} else {
  // something
}
```

src/components/expression/expression-renderer.js
Shauro: Good idea.

src/components/expression/expression-renderer.js
yubowenok: If you really need to divide by 3, then probably you want to create another constant called TEXT_HEIGHT_OVER_THREE (or some other better name :) ) and then declare it in the prototype. In the comments of this variable you might explain why you are using this "dividing by 3" value.

``` js
/**
 * A third of text height. A third of the text height is really important because ...
 * @const {number}
 */
genotet.ExpressionRenderer.TEXT_HEIGHT_OVER_THREE = ...
```

src/components/expression/expression-renderer.js
liana-yang, line 305: But its parent svg is appened in js. We can write like this and load it as template. But I think it is not that necessary...

src/components/expression/expression-renderer.js
liana-yang: It is more than 80 characters in this line....

src/components/expression/expression-renderer.js
yubowenok: https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting

or if this is still not convincing enough, then your favorite airbnb style guide too:

https://github.com/airbnb/javascript#16.2

:)

src/components/expression/expression-renderer.js
yubowenok: Okay I just opened a text editor and tried it. Please ignore my noise ;)

src/components/expression/expression-renderer.js
yubowenok, line 305: Sure. Up to you.

src/components/expression/expression-renderer.js
``` js
// good
if (condition1 && condition2) {
}

// good
if (veryVeryLongCondition1 &&
  veryVeryLongCondition2) {
}

// good
if (halfOfSomeVeryLongExpression +
  theOtherHalfOfSomeVeryLongExpression) {
}

// bad
if ( // what a lonely bracket
  veryVeryLongCondition1 &&
  veryVeryLongCondition2
) {
}

// bad
if ( // what a lonely bracket...
  halfOfSomeVeryLongExpression +
  theOtherHalfOfSomeVeryLongExpression
) {
}


```

src/components/expression/expression-renderer.js
```diff
@@ -6,7 +6,7 @@
 
 /**
  * ExpressionRenderer renders the visualizations for the ExpressionView.
- * @param {!jQuery} container View container.
+ * @param {!d3.selection} container View container.
  * @param {!Object} data Data object to be written.
  * @extends {ViewRenderer}
  * @constructor
```
yubowenok, line 12: Now that you typedef-ed Profile, you can use your defined type for this.

src/components/expression/expression-renderer.js
```diff
@@ -465,22 +604,51 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
       this.PROFILE_MARGINS_.LEFT = this.DEFAULT_PROFILE_MARGIN;
     }
     else {
-      this.PROFILE_MARGINS_.LEFT = this.geneLabelWidth_ + this.LABEL_MARGIN_;
+      this.PROFILE_MARGINS_.LEFT = this.geneLabelWidth_ + this.LABEL_MARGIN;
```
yubowenok: You are changing a constant variable, which isn't allowed though we are not blocking it in compilation.

src/components/expression/expression-renderer.js
```diff
@@ -518,34 +687,45 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
-        var value = heatmapData.values[geneIndex][conditionIndex];
-        _(this.data.profiles[i]).extend({
-          container_: d3.event.target,
-          hoverColumn: conditionIndex,
-          hoverConditionName: heatmapData.conditionNames[conditionIndex],
-          hoverValue: value
-        });
+          xScale.invert(d3.mouse(d3.event.target)[0]) + 0.5
+        );
+        var value = heatmapData.values[profile.row][conditionIndex];
```
yubowenok: why did you change this back to not using extend?

src/components/expression/expression-renderer.js
```diff
@@ -553,17 +733,25 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
-          container_: d3.event.target,
-          hoverColumn: conditionIndex,
-          hoverConditionName: heatmapData.conditionNames[conditionIndex],
-          hoverValue: value
-        });
+        var conditionIndex = Math.floor(
+          xScale.invert(d3.mouse(d3.event.target)[0]) + 0.5
+        );
+        var value = heatmapData.values[profile.row][conditionIndex];
+        this.data.profiles[i].container_ = d3.event.target;
```
yubowenok: ditto

src/panel-manager.js
```diff
@@ -169,3 +186,35 @@ genotet.panelManager.closeAllPanels = function() {
...
+   * Exclude the initial invisible panel.
+   */
+  var newTabHeight = Math.floor((tabContentHeight -
+    genotet.panelManager.TAB_HEIGHT_DIFFERENCE_ -
+    genotet.panelManager.TOGGLE_BTN_HEIGHT_) / tabCount - 1
+  );
+
+  tabSelector
+    .find('li')
+    .css('margin-bottom', (
```
yubowenok: pls fix style

src/panel-manager.js
```diff
@@ -169,3 +186,35 @@ genotet.panelManager.closeAllPanels = function() {
...
+      newTabHeight - genotet.panelManager.TAB_MARGIN_BOTTOM_DIFFERENCE_) + 'px'
+    );
+  tabSelector
+    .find('li:nth-child(2) a')
+    .css('margin-top', (
```
yubowenok: ditto

src/components/expression/expression-renderer.js
```diff
@@ -553,17 +733,25 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+        this.data.profiles[i].hoverConditionName =
+          heatmapData.conditionNames[conditionIndex];
+        this.data.profiles[i].hoverValue = value;
+        this.data.profiles[i].clicked = true;
         this.signal('pathClick', this.data.profiles[i]);
       }.bind(this));
+    if (this.data.profiles[i].clicked) {
+      this.highlightLabelsForClickedProfile_(this.data.profiles[i]);
+    }
+    else {
```
yubowenok: pls fix style

src/components/expression/expression-renderer.js
```diff
@@ -465,22 +604,51 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
     else {
-      this.PROFILE_MARGINS_.LEFT = this.geneLabelWidth_ + this.LABEL_MARGIN_;
+      this.PROFILE_MARGINS_.LEFT = this.geneLabelWidth_ + this.LABEL_MARGIN;
     }
     this.svgProfile_.selectAll('*').attr('display', 'inline');
   }
 
   var heatmapData = this.data.matrix;
-  this.svgProfile_.selectAll('g').remove();
+  this.svgProfile_.selectAll('*').remove();
```
yubowenok: not using d3 update?

src/components/expression/expression-renderer.js
```diff
@@ -465,22 +604,51 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+  var i = 0;
+  var profileCount = this.data.profiles.length;
+  while (i < profileCount) {
+    var geneIndex = heatmapData.geneNames
+      .indexOf(this.data.profiles[i].geneName);
+    if (geneIndex == -1) {
+      this.data.profiles.splice(i, 1);
+      profileCount--;
+    }
+    else {
```
yubowenok: fix style pls

src/components/expression/expression-renderer.js
```diff
@@ -465,22 +604,51 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+      this.DEFAULT_PROFILE_LEGEND_MARGIN
+    ]));
+  var legendHeight = this.DEFAULT_PROFILE_LEGEND_HEIGHT -
+    this.DEFAULT_PROFILE_LEGEND_MARGIN;
+
+  var i = 0;
+  var profileCount = this.data.profiles.length;
+  while (i < profileCount) {
+    var geneIndex = heatmapData.geneNames
+      .indexOf(this.data.profiles[i].geneName);
```
yubowenok: Usually we write chaining methods in new lines. Since this is not chaining method, I think it's better to be

``` js
var geneIndex = heatmapData.geneNames.indexOf(
  this.data.profiles[i].geneName); 
```

src/components/expression/expression-renderer.js
```diff
@@ -488,17 +656,18 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
     .append('svg:g').call(yAxis)
     .classed('axis', true)
-    .attr('transform', 'translate(' + this.PROFILE_MARGINS_.LEFT + ', 0)');
+    .attr('transform', genotet.utils.getTransform([
+      this.PROFILE_MARGINS_.LEFT,
+      0
+    ]));
 
   var profileContent = this.svgProfile_.append('g')
     .attr(
```
yubowenok: could you pls fix the styles here too?

src/components/expression/expression.js
```diff
@@ -73,8 +73,11 @@ genotet.ExpressionView = function(viewName, params) {
     }.bind(this))
     .on('genotet.cellClick', function(event, cell) {
       this.renderer.highlightLabelsForClickedCell_(cell);
-      this.panel.displayCellInfo_(cell.geneName, cell.conditionName,
-        cell.value);
+      this.panel.displayCellInfo_(cell.geneName,
```
yubowenok: why moving cell.conditionName to a second line? looks like it can fit in this line

src/components/expression/expression.js
```diff
@@ -90,8 +93,11 @@ genotet.ExpressionView = function(viewName, params) {
     }.bind(this))
     .on('genotet.pathClick', function(event, profile) {
       this.renderer.highlightLabelsForClickedProfile_(profile);
-      this.panel.displayPathInfo_(profile.geneName, profile.hoverConditionName,
-        profile.hoverValue);
+      this.panel.displayPathInfo_(profile.geneName,
+        profile.hoverConditionName, profile.hoverValue);
```
yubowenok: ditto

src/components/expression/expression-renderer.js
liana-yang, line 12: I don't understand this. These params are not profiles.

src/panel-manager.js
```diff
@@ -5,10 +5,22 @@
 /** @const */
 genotet.panelManager = {};
 
-/** @const {number} */
-genotet.panelManager.COLLAPSED_WIDTH = 32;
-/** @const {number} */
-genotet.panelManager.TRANSITION_TIME = 300;
+/** @private @const {number} */
+genotet.panelManager.COLLAPSED_WIDTH_ = 32;
```
yubowenok: genotet.panelManager.COLLAPSED_WIDTH is used in viewManager (line 23) and needs to be public variable.

yubowenok: It is also used in binding.js (line 103)

src/components/expression/expression-renderer.js
yubowenok, line 12: misplaced my comment...

src/components/expression/expression-renderer.js
```diff
@@ -20,40 +20,16 @@ genotet.ExpressionRenderer = function(container, data) {
...
    *   row: number,
-   *   color: string
+   *   hoverColumn: number,
+   *   hoverConditionName: string,
+   *   hoverValue: number,
+   *   color: string,
+   *   clicked: bool
    * }>}
    */
   this.data.profiles;
```
yubowenok, line 21: "Now that you typedef-ed Profile, you can use your defined type for this." 
Should this be !Array<!genotet.ExpressionRenderer.Profile>?

src/panel-manager.js
```diff
@@ -169,3 +186,33 @@ genotet.panelManager.closeAllPanels = function() {
...
+   */
+  var newTabHeight = Math.floor((tabContentHeight -
+    genotet.panelManager.TAB_HEIGHT_DIFFERENCE_ -
+    genotet.panelManager.TOGGLE_BTN_HEIGHT_) / tabCount - 1
+  );
+
+  tabSelector
+    .find('li')
+    .css('margin-bottom', (newTabHeight -
+      genotet.panelManager.TAB_MARGIN_BOTTOM_DIFFERENCE_) + 'px');
```
yubowenok: I think you don't need "+ 'px'". jQuery by default thinks they are in pixels.

src/components/expression/expression-renderer.js
liana-yang, line 12: = =|||

src/components/expression/expression-renderer.js
```diff
@@ -518,28 +690,44 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
     })
     .y(yScale)
     .interpolate('linear');
+
+  this.svgLegend_.selectAll('rect').remove();
+  this.svgLegend_.selectAll('text').remove();
+  this.profileContent_.selectAll('path').remove();
```
yubowenok: d3 update TBD

src/components/expression/expression-renderer.js
```diff
@@ -74,99 +35,139 @@ genotet.ExpressionRenderer = function(container, data) {
...
+ *   container_: d3.selection,
+ *   geneName: string,
+ *   conditionName: string,
+ *   row: number,
+ *   column: number,
+ *   value: number,
+ *   colorscaleValue: string
+ * }
+ */
+genotet.ExpressionRenderer.Cell = {
```
yubowenok: typedef doesn't need value assignment

``` js
genotet.ExpressionRenderer.Cell;
```

src/components/expression/expression-renderer.js
```diff
@@ -74,99 +35,139 @@ genotet.ExpressionRenderer = function(container, data) {
...
+ *   geneName: string,
+ *   row: number,
+ *   hoverColumn: number,
+ *   hoverConditionName: string,
+ *   hoverValue: number,
+ *   color: string,
+ *   clicked: bool
+ * }
+ */
+genotet.ExpressionRenderer.Profile = {
```
yubowenok: ditto

src/components/expression/expression-renderer.js
```diff
@@ -459,58 +615,73 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+      0,
+      this.profileHeight_ - this.profileMargins_.bottom
+    ]));
+  this.svgProfileYAxis_.call(yAxis)
+    .attr('transform', genotet.utils.getTransform([
+      this.profileMargins_.left,
+      0
+    ]));
+
+  this.profileContent_
```
yubowenok: Why not computing sizes in layout()?

src/components/expression/expression-renderer.js
```diff
@@ -518,53 +689,103 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+  legendRect
+    .attr('height', legendHeight)
+    .attr('width', legendHeight)
+    .attr('x', function(profile, i) {
+      return i * (legendHeight +
+        profile.geneName.length * this.GENE_LABEL_WIDTH_FACTOR);
+    }.bind(this))
+    .style('fill', function(profile) {
+      return this.COLOR_CATEGORY[genotet.utils.hashString(profile.geneName) %
+      this.COLOR_CATEGORY_SIZE];
```
yubowenok: Please indent here.

src/components/expression/expression-renderer.js
```diff
@@ -518,53 +689,103 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+  profilePath
+    .classed('profile', true)
+    .attr('d', function(profile) {
+      return line(heatmapData.values[profile.row]);
+    })
+    .attr('transform', genotet.utils.getTransform([
+      this.heatmapWidth_ / (heatmapData.conditionNames.length * 2),
+      0
+    ]))
+    .attr('stroke', function(profile, i) {
```
yubowenok: Use style to set stroke.

src/components/expression/expression-renderer.js
```diff
@@ -518,53 +689,103 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+      var pathColor = this.COLOR_CATEGORY[genotet.utils.hashString(
+        profile.geneName) % this.COLOR_CATEGORY_SIZE];
+      this.data.profiles[i].color = pathColor;
+      return pathColor;
+    }.bind(this))
+    .attr('fill', 'none')
```
yubowenok: If a property applies to every element, then declare it in css.

One exception is attrs. However fill is style...

src/components/expression/expression-renderer.js
```diff
@@ -673,23 +895,68 @@ genotet.ExpressionRenderer.prototype.highlightLabelsForClickedCell_ = function(
...
-      (profile.row + 1) + ')')
-    .attr('fill', profile.color);
-  this.svgConditionLabels_.select('text:nth-child(' +
-      (profile.hoverColumn + 1) + ')')
-    .attr('fill', profile.color);
-};
+  function(profile) {
+    this.unhighlightLabelsForClickedProfile_();
+    this.svgGeneLabels_.select('text:nth-child(' + (profile.row + 1) + ')')
+      .attr('fill', profile.color)
```
yubowenok: style for fill and stroke.

src/components/expression/expression-renderer.js
```diff
@@ -673,23 +895,68 @@ genotet.ExpressionRenderer.prototype.highlightLabelsForClickedCell_ = function(
...
+  };
+
+/**
+ * Unhighlights all the labels of profiles.
+ * @private
+ */
+genotet.ExpressionRenderer.prototype.unhighlightLabelsForClickedProfile_ =
+  function() {
+    this.svgGeneLabels_.selectAll('text')
+      .attr('fill', 'black')
```
yubowenok: ditto

src/panel-manager.js
```diff
@@ -7,8 +7,20 @@ genotet.panelManager = {};
...
 genotet.panelManager.COLLAPSED_WIDTH = 32;
-/** @const {number} */
-genotet.panelManager.TRANSITION_TIME = 300;
+/** @private @const {number} */
+genotet.panelManager.TRANSITION_TIME_ = 300;
+/** @private @const {number} */
+genotet.panelManager.TOGGLE_BTN_HEIGHT_ = 50;
+
+/**
+ * The difference between the width and other attributes of the tabs container.
```
yubowenok: If it's not JSDoc you don't need /*\* ... */ comments. Just // will be fine.

Besides this explanation is still very confusing. I cannot get it at all ?_?...

src/components/expression/expression-renderer.js
liana-yang: As I known, style and attr can both set fill and stroke attributes. When they are set to an element at  same time, style takes the priority. I used attr here because I used attr to set attributes while clicking the gene profile and used style to set attributes while clicking the heatmap cell to ensure that clicking the heatmap cell takes the priority when you clicked the same position of gene profile before.

src/components/expression/expression-renderer.js
liana-yang: ditto

src/panel-manager.js
liana-yang: What is confusing about this comment?

yubowenok: What is "the difference between the width and other attributes"? This is a very confusing sentence. I see no clue how to compare "width" and "other attribute", thus don't know what "difference" there is.

src/components/expression/expression-renderer.js
yubowenok: According to [W3C recommendations](http://www.w3.org/TR/SVG/propidx.html) fill is a _property_. We consistently use style to set _properties_  to distinguish them from non-property attributes such as "r", "cx", etc.

Your way of using both attr and style doesn't make much sense to me as you are giving a same element two rendering styles at the same time. The two styles are competing. Competing styles should be avoided as much as possible, e.g. in CSS you want to avoid [Specifity Wars](https://stuffandnonsense.co.uk/archives/css_specificity_wars.html).

If you want to highlight interacted elements you can simply add classes or use ":state" selector.

src/components/expression/expression-renderer.js
```diff
@@ -137,36 +54,103 @@ genotet.ExpressionRenderer = function(container, data) {
   this.heatmapHeight_ = 0;
 
   /**
-   * Margin of the gene profile.
-   * @private {number}
+   * Margins of the gene profile.
+   * @private {!Object}
```
yubowenok: Better be specific: @private {!Object<number>}. Same below.

src/components/expression/expression-renderer.js
```diff
@@ -137,36 +54,103 @@ genotet.ExpressionRenderer = function(container, data) {
...
+    BOTTOM: 5,
+    LEFT: 40
+  };
 };
 
 genotet.utils.inherit(genotet.ExpressionRenderer, genotet.ViewRenderer);
 
+/**
+ * Cell object storing the rendering properties of expression cell.
+ * @typedef {
```
yubowenok: Missing one level of curly bracket - you are declaring an object type.

``` js
/**
 * @typedef {{
 * ...
 * }}
 */
```

src/components/expression/expression-renderer.js
```diff
@@ -310,15 +373,33 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+    transformLeft += this.LABEL_MARGIN;
+  } else if (this.data.options.showProfiles) {
     transformLeft += this.DEFAULT_PROFILE_MARGIN;
   }
   if (this.data.options.showConditionLabels) {
-    transformTop += this.LABEL_MARGIN_;
+    transformTop += this.conditionLabelHeight_ +
+      this.LABEL_MARGIN + this.LABEL_DIFFERENCE;
+  }
+
```
yubowenok: transformLeft and transformTop are applied to every heatmapRow. Why not applying them to this.svgHeatmapContent_ instead? Also this could be written in layout().

src/components/expression/expression-renderer.js
```diff
@@ -342,7 +423,7 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
     .style('fill', colorScale)
     .on('mouseover', function(value, i, j) {
       var hoverCell = d3.event.target;
-      var cell = this.cell_ = {
+      var cell = genotet.ExpressionRenderer.Cell = {
```
yubowenok: Wrong usage. You should not be using writing typedef-ed class name anywhere other than your declaraion.

Also what you do here is dangerous. You are passing object reference to a new object. What's the purpose? If you want to initialize it to empty, then you should create a constructor class for this object.

``` js
genotet.ExpressionRenderer.Cell = function() {
// ...
};
var cell = new genotet.Expressionrenderer.Cell();
_(cell).extend({
  // ...
});
```

src/components/expression/expression-renderer.js
```diff
@@ -442,7 +522,7 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
   labels
     .attr('transform', genotet.utils.getTransform([
```
yubowenok: When you want to apply a transform to every element under a group, apply to the group instead.

src/components/expression/expression-renderer.js
```diff
@@ -452,6 +532,71 @@ genotet.ExpressionRenderer.prototype.drawMatrixConditionLabels_ = function() {
...
+  } else if (this.data.options.showProfiles) {
+    transformLeft += this.DEFAULT_PROFILE_MARGIN;
+  }
+  if (this.data.options.showConditionLabels) {
+    transformTop += this.conditionLabelHeight_ +
+      this.LABEL_MARGIN + this.LABEL_DIFFERENCE;
+  }
+  this.svgHeatmapGradient_
+    .attr('transform',
+      genotet.utils.getTransform([transformLeft, transformTop]));
```
yubowenok: Please layout().

src/components/expression/expression-renderer.js
liana-yang: It is @private {!Object}! The red lines are deleted..

src/components/expression/expression-renderer.js
yubowenok: If your layout() becomes too complicated, then define multiple methods like

``` js
genotet.ExpressionRenderer.layoutGradient_ = function() {
}
```

src/components/expression/expression-renderer.js
yubowenok: I mean use

``` js
@private {!Object<number>}
```

instead of vague

``` js
@private {!Object}
```

src/components/expression/expression-renderer.js
liana-yang: I cannot add classes simply here because the color of the path need to be calculated. They are not constants.

yubowenok: Then you may want to handle it in other way. Here it should be

``` js
.style('fill', profile.color)
```

src/components/expression/expression-renderer.js
yubowenok: Okay looks like markdown will eat angular bracketed elements...

```
@private {!Object<number>}
```

is shown as @private {!Object} outside code block.

src/components/expression/expression-renderer.js
```diff
@@ -184,6 +168,34 @@ genotet.ExpressionRenderer.prototype.initLayout = function() {
...
+   * @private {!d3.selection}
+   */
+  this.svgLegend_ = this.svgProfile_.append('g')
+    .classed('legend', true);
+
+  /**
+   * SVG group for profile x axis.
+   * @private {!d3.selection}
+   */
+  this.svgProfileXAxis_ = this.svgProfile_.append('svg:g')
```
yubowenok: what's the purpose of "svg:g"?

src/components/expression/expression-renderer.js
```diff
@@ -459,58 +604,67 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
 
+  this.svgLegend_
+    .attr('transform', genotet.utils.getTransform([
+      this.profileMargins_.left,
+      this.DEFAULT_PROFILE_LEGEND_MARGIN
+    ]));
+  var legendHeight = this.DEFAULT_PROFILE_LEGEND_HEIGHT -
+    this.DEFAULT_PROFILE_LEGEND_MARGIN;
+
+  var i = 0;
```
yubowenok: This block of code is a pre-processing step that does not need to be repeated on redraw. You can compute this whenever the data is changed.

Besides, indexOf is inefficient. A better way could be

``` js
// In utils.js
/**
 * Converts an array of string or object with string keys to a set with the strings
 * as keys. The values will be set to all true.
 * @param {!Array<string>|!Object<*>} collection
 * @return {!Object<boolean>}
 */
genotet.utils.keySet = function(collection) {
  var obj = {};
  if (collection instanceof Array) {
    collection.forEach(function(element) {
      obj[element] = true;
    });
  } else {
    for (var key in collection) {
      obj[key] = true;
    }
  }
  return obj;
};

// Here
var geneNames = genotet.utils.keySet(heatmapData.geneNames);
this.data.profiles = this.data.profiles.filter(function(profile) {
  return profile.geneName in geneNames;
});
```

src/panel-manager.js
```diff
@@ -169,3 +183,33 @@ genotet.panelManager.closeAllPanels = function() {
...
+ * Adjust tab height.
+ */
+genotet.panelManager.adjustTabHeight = function() {
+  var tabSelector = $('.sideways');
+  var tabCount = tabSelector.find('li').length - 1;
+  var tabHeight = tabSelector.find('li:nth-child(2) a').outerWidth();
+  var tabContentHeight = $('.tab-content').outerHeight();
+
+  /**
+   * Exclude the initial invisible panel.
```
yubowenok:
``` js
// Exclude the initial invisible panel by deducting 1.
```

Use // if comment is not JSDoc.

src/components/expression/expression-renderer.js
```diff
@@ -74,99 +35,138 @@ genotet.ExpressionRenderer = function(container, data) {
...
+    BOTTOM: 5,
     LEFT: 40
   };
 
   /**
-   * Height of the gene profile plot.
-   * This value will be zero when profiles are not shown.
-   * @private {number}
+   * Clicked object for heatmap cells and gene profiles.
+   * @private {!Object}
```
yubowenok: Should this be genotet.ExpressionRenderer.Cell?

src/components/expression/expression-renderer.js
```diff
@@ -459,58 +620,53 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+      this.profileMargins_.left = this.geneLabelWidth_ + this.LABEL_MARGIN;
     }
     this.svgProfile_.selectAll('*').attr('display', 'inline');
   }
 
   var heatmapData = this.data.matrix;
-  this.svgProfile_.selectAll('g').remove();
   this.svgProfile_.attr('width', this.canvasWidth_);
 
+  this.svgLegend_
```
yubowenok: Setting positions of svg groups could be moved to layout(), so does the computation of group translation, e.g. this.profileMargins_.left here.

src/utils.js
```diff
@@ -266,6 +266,27 @@ genotet.utils.swap = function(arr, i, j) {
 };
 
 /**
+ * Converts an array of string or object with string keys to a set
+ * with the strings as keys. The values will be set to all true.
+ * @param {!Array<string>|!Object<*>} collection
+ * @return {!Object<boolean>}
+ * @private
+ */
+genotet.utils.keySet_ = function(collection) {
```
yubowenok: Make this public, or expose it via underscore.

If latter, add the following to utils.js.

``` js
/**
 * Initializes the utils. Creates underscore mixins.
 */
genotet.utils.init = function() {
  _.mixin({
    keySet: genotet.utils.keySet_,
  });
};
```

and then you can do

``` js
var myArrayDict = _(myArray).keySet();
```

Don't forget to call genotet.utils.init() of course.

src/components/expression/expression-renderer.js
```diff
@@ -309,70 +454,60 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
         container_: hoverCell,
         colorscaleValue: colorScale(value)
-      };
+      });
       this.signal('cellUnhover', cell);
     }.bind(this))
     .on('click', function(value, i, j) {
       var hoverCell = d3.event.target;
-      var cell = this.cell_ = {
+      _(this.clickedObject_).extend({
```
yubowenok, line 587: Is this a Cell? Why not newing a cell using constructor?

src/components/expression/expression-renderer.js
```diff
@@ -220,29 +293,89 @@ genotet.ExpressionRenderer.prototype.layout = function() {
...
+  if (this.data.options.showGeneLabels) {
+    heatmapGradientTransformLeft += this.geneLabelWidth_ + this.LABEL_MARGIN;
+  } else if (this.data.options.showProfiles) {
+    heatmapGradientTransformLeft += this.DEFAULT_PROFILE_MARGIN;
+  }
+  if (this.data.options.showConditionLabels) {
+    heatmapGradientTransformTop += this.conditionLabelHeight_ +
+      this.LABEL_MARGIN + this.LABEL_DIFFERENCE;
+  }
+  this.svgHeatmapGradient_
```
``` js
this.svgHeatmapGradient_.attr('transform', genotet.utils.getTransform([
  heatmapGradientTransformLeft,
  heatmapGradientTransformTop
]));
```

src/components/expression/expression-renderer.js
```diff
@@ -309,70 +454,60 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
-    })
+      return j * this.cellHeight;
+    }.bind(this))
     .style('stroke', colorScale)
     .style('fill', colorScale)
     .on('mouseover', function(value, i, j) {
       var hoverCell = d3.event.target;
-      var cell = this.cell_ = {
+      var cell = new genotet.ExpressionRenderer.Cell();
+      _(cell).extend({
```
yubowenok: Umm if you are extending many attributes, probably a better way is to construct by params.

``` js
genotet.ExpressionRenderer.Cell = function(params) {
  this.container = params.container;
  this.geneName = params.geneName;
  // ... and so on
}

// Here:
var cell = new genotet.ExpressionRenderer.Cell({
  container: hoverCell,
  geneName: heatmapData.geneNames[j],
  // ... and so on
});
```

src/components/expression/expression-renderer.js
```diff
@@ -518,53 +674,93 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+      _(this.data.profiles[i]).extend({
+        container_: d3.event.target,
+        hoverColumn: conditionIndex,
+        hoverConditionName: heatmapData.conditionNames[conditionIndex],
+        hoverValue: value
+      });
+      this.signal('pathHover', this.data.profiles[i]);
+    }.bind(this))
+    .on('mouseout', function(profile, i) {
+      this.data.profiles[i].container_ = d3.event.target;
```
yubowenok: You shouldn't be accessing a private variable like this. Either make it public or create a wrapper set function.

``` js
genotet.ExpressionRenderer.Profile.prototype.setContainer = function(container) {
  this.container_ = container;
};
```

src/components/expression/expression-renderer.js
```diff
@@ -518,53 +674,93 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
+      this.signal('pathUnhover', this.data.profiles[i]);
+    }.bind(this))
+    .on('click', function(profile, i) {
+      var conditionIndex = Math.floor(
+        xScale.invert(d3.mouse(d3.event.target)[0]) + 0.5);
+      var geneIndex = heatmapData.geneNames.indexOf(
+        this.data.profiles[i].geneName);
+      var value = heatmapData.values[geneIndex][conditionIndex];
+      _(this.clickedObject_).extend({
```
yubowenok, line 822: ditto. Maybe new a Cell object by params?

src/components/expression/expression-renderer.js
Shauro: Can it initiate?

src/components/expression/expression-renderer.js
liana-yang, line 587: Fixed.

src/components/expression/expression-renderer.js
liana-yang, line 822: Here is not creating a new Cell object, but updating the attributes of this.clickedObject_.

src/components/expression/expression-renderer.js
```diff
@@ -74,99 +35,138 @@ genotet.ExpressionRenderer = function(container, data) {
...
+ * Cell object storing the rendering properties of expression cell.
+ * @struct
+ * @constructor
+ * @param {!Object} params
+ */
+genotet.ExpressionRenderer.Cell = function(params) {
+  this.container = params.container;
+  this.geneName = params.geneName;
+  this.conditionName = params.conditionName;
+  this.row = params.row;
```
yubowenok: If you have default values you can say 

``` js
this.row = params.row != null ? params.row : -1;
```

so that you don't need to pass in default values for every constructed objects.

src/components/expression/expression-renderer.js
```diff
@@ -74,99 +35,138 @@ genotet.ExpressionRenderer = function(container, data) {
...
-  /**
-   * Margin of the gene profile.
-   * @private {number}
-   */
-  this.geneProfileMargin_ = 0;
+/**
+ * Cell object storing the rendering properties of expression cell.
+ * @struct
+ * @constructor
+ * @param {!Object} params
```
``` js
@param {!{
  container: !d3.selection,
  geneName: string,
  conditionName: string,
  row: number=,
  column: number=,
  value: number=,
  colorscaleValue: ??? // I'm not sure
}} params
  container: {explanation of container}
  geneName: {explanation of geneName}
  .. and so on
```

src/components/expression/expression-renderer.js
```diff
@@ -220,27 +293,110 @@ genotet.ExpressionRenderer.prototype.layout = function() {
...
-    this.conditionLabelHeight_;
+  this.heatmapHeight_ = this.canvasHeight_ -
+    this.profileHeight_ - this.DEFAULT_HEATMAP_GRADIENT_HEIGHT;
   if (this.data.options.showConditionLabels) {
-    this.heatmapHeight_ -= this.LABEL_MARGIN_;
+    this.heatmapHeight_ -= this.conditionLabelHeight_ +
+      this.LABEL_MARGIN + this.LABEL_DIFFERENCE;
   }
 
+  var heatmapData = this.data.matrix;
```
yubowenok, line 370: The layout() function is a bit too bulky now. Consider modularizing it and creating sub-functions for sub-routines (maybe later).

src/components/expression/expression-renderer.js
yubowenok, line 822: ok

src/components/expression/expression-renderer.js
```diff
@@ -74,99 +35,161 @@ genotet.ExpressionRenderer = function(container, data) {
...
-   * Margin of the gene profile.
-   * @private {number}
-   */
-  this.geneProfileMargin_ = 0;
+/**
+ * Cell object storing the rendering properties of expression cell.
+ * @struct
+ * @constructor
+ * @param {!{
+ * container: !d3.selection,
```
yubowenok: Please correct indentation.

src/components/expression/expression-renderer.js
```diff
@@ -74,99 +35,161 @@ genotet.ExpressionRenderer = function(container, data) {
...
+ * geneName: string,
+ * conditionName: string,
+ * row: number,
+ * column: number,
+ * value: number,
+ * colorscaleValue: string
+ * }} params
+ * container: {Container of the expression cell}
```
yubowenok: indent 2 spaces for all lines starting here. you don't need curly brackets for these lines.

src/components/expression/expression-renderer.js
```diff
@@ -74,99 +35,161 @@ genotet.ExpressionRenderer = function(container, data) {
...
+ * geneName: {Gene Name of the expression cell}
+ * conditionName: {Condition Name of the expression cell},
+ * row: {Row index of the expression cell},
+ * column: {Column index of the expression cell},
+ * value: {Value of the expression cell},
+ * colorscaleValue: {Colorscale value of the expression cell}
+ */
+genotet.ExpressionRenderer.Cell = function(params) {
+  this.container = params != null ? params.container : null;
```
yubowenok: Add JSDoc for these variables.

``` js
/** @type {!jQuery} */
this.container = ...


// or more verbose
/**
 * Container of the cell blahblah.
 * @type {!d3.selection}
 */
this.container = params.container;
```

Besides as you specify non-nullable params, you don't need to check if params is null. Otherwise, you should say in the JSDoc that params is nullable.

src/components/expression/expression-renderer.js
```diff
@@ -74,99 +35,161 @@ genotet.ExpressionRenderer = function(container, data) {
...
-  this.geneProfileMargin_ = 0;
+/**
+ * Cell object storing the rendering properties of expression cell.
+ * @struct
+ * @constructor
+ * @param {!{
+ * container: !d3.selection,
+ * geneName: string,
+ * conditionName: string,
+ * row: number,
```
yubowenok: if a variable has default value, then it should be number= instead of number

src/components/expression/expression-renderer.js
```diff
@@ -74,99 +35,161 @@ genotet.ExpressionRenderer = function(container, data) {
...
+ * row: {Row index of the expression cell},
+ * column: {Column index of the expression cell},
+ * value: {Value of the expression cell},
+ * colorscaleValue: {Colorscale value of the expression cell}
+ */
+genotet.ExpressionRenderer.Cell = function(params) {
+  this.container = params != null ? params.container : null;
+  this.geneName = params != null ? params.geneName : null;
+  this.conditionName = params != null ? params.conditionName : null;
+  this.row = params != null ? params.row : -1;
```
yubowenok: If you are setting default value for this.row, it should be

``` js
this.row = params.row != null ? params.row : -1;
```

yubowenok:
LGTM. You may now merge.

PR #52, opened 2016-01-06
Expression Matrix: Box zoom and matrix row/column manipulation.

liana-yang:
Added TFA profiles in expression matrix.


server/expression.js
```diff
@@ -108,6 +108,7 @@ module.exports = {
    * @param {string} fileExp File name of the expression matrix.
```
yubowenok: Server code will change after merging #54 . Please adjust accordingly after #54 is merged.

src/components/expression/expression-loader.js
```diff
@@ -31,6 +33,7 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
 genotet.ExpressionLoader.prototype.load = function(matrixName, geneRegex,
                                                    conditionRegex) {
   this.loadExpressionMatrix_(matrixName, geneRegex, conditionRegex);
+  this.loadExpressionTfaProfile_('b-subtilis', 'sigA', conditionRegex);
```
yubowenok: Constant doesn't seem to be right here.

src/components/expression/expression-loader.js
```diff
@@ -67,6 +70,34 @@ genotet.ExpressionLoader.prototype.loadExpressionMatrix_ = function(matrixName,
 };
 
 /**
+ * Implements the expression matrix TFA profile loading ajax call.
+ * Since the matrix may contain a large number of entries, we use POST request.
```
yubowenok: Comments do not correctly reflect implementation... as the request is actually GET.

src/components/expression/expression-panel.html
```diff
@@ -30,6 +30,10 @@
       <button id="removeCondition" class="btn btn-default btn-xs glyphicon glyphicon-minus" title="remove conditions"></button>
     </div>
   </div>
+  <div id="zoom-out" class="panel-section">
+    <div class="item-title">Zoom Out</div>
+    <button id="out" class="zoom-out btn btn-default btn-xs glyphicon glyphicon-minus"></button>
```
yubowenok: Probably too many ids than necessary. Ids are useful unique identifiers but don't use too many of them. In this case you can just keep zoom-out and then use #zoom-out > button.

src/components/expression/expression-renderer.js
```diff
@@ -178,6 +178,42 @@ genotet.ExpressionRenderer.Profile = function(params) {
   this.color = params.color != null ? params.color : null;
 };
 
+/**
+ * Zoom status object storing the status of expression matrix.
+ * @param {!{
```
yubowenok: My bad here. I double-checked the docs and see that record types and typedef (usually seen as {{...}}) is non-nullable by default. We can get rid of the exclamation mark.

src/components/expression/expression-renderer.js
```diff
@@ -594,7 +633,119 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+        d3.selectAll('.label-selected').classed('label-selected', false);
+
+        rowStart = Math.floor(selectedRange.y / cellHeight);
+        rowEnd = Math.floor((selectedRange.y + selectedRange.height) /
+          cellHeight);
+        columnStart = Math.floor(selectedRange.x / cellWidth);
+        columnEnd = Math.floor((selectedRange.x + selectedRange.width) /
+          cellWidth);
+        for (var i = rowStart; i <= rowEnd; i++) {
+          d3.select('.gene-labels text:nth-of-type(' + (i + 1) + ')')
```
yubowenok: Seems dangerous and inefficient. How do you ensure the order of the d3 appended elements so that nth-of-type always works properly? What happens if some of the elements are removed and updated? If you always draw the elements based on their order in the data array, then that's fine.

It would be better to do all the selections in one batch instead of selecting one by one. I think it may be even more efficient to selectAll and then 

``` js
.classed('label-selected', function(cell, index) {
  return index >= columnStart && index <= columnEnd;
}
```

, as opposed to d3.select each with nth-of-type.

src/components/expression/expression-renderer.js
```diff
@@ -968,13 +1118,40 @@ genotet.ExpressionRenderer.prototype.highlightLabelsAfterUpdateData_ =
       this.clickedObject_.column = heatmapData.conditionNames.indexOf(
         this.clickedObject_.conditionName
       );
-      this.signal('expressionClick', this.clickedObject_);
+    }
+    if (this.clickedObject_.row != -1 && this.clickedObject_.column != -1) {
```
yubowenok: If you are using a lot of -1, maybe define a constant.

src/components/expression/expression-renderer.js
```diff
@@ -968,13 +1118,40 @@ genotet.ExpressionRenderer.prototype.highlightLabelsAfterUpdateData_ =
...
+        this.signal('expressionClick', this.clickedObject_);
     } else {
       this.signal('expressionUnclick');
     }
   };
 
 /**
+ * Load expression matrix data after zoom in and out the heatmap.
+ * @param {!Object} params Parameters for expression zoom in.
```
yubowenok: Please be specific about Object. It is encouraged to write object attributes explicitly.

src/components/expression/expression-renderer.js
```diff
@@ -178,6 +178,42 @@ genotet.ExpressionRenderer.Profile = function(params) {
...
 
+/**
+ * Zoom status object storing the status of expression matrix.
+ * @param {!{
+ *   matrixName: (?string|undefined),
+ *   geneRegex: (?string|undefined),
+ *   conditionRegex: (?string|undefined),
+ *   geneNames: (?string|undefined),
+ *   conditionNames: (?string|undefined)
+ * }} params
```
yubowenok, line 207: Everything can be undefined is a bit too permissive. Please be as strict as possible. It seems that matrixName is never undefined in your usage, is it?

src/components/expression/expression-renderer.js
yubowenok: Alternatively, you can assign a unique id to every cell and efficiently select every element using their ids.

src/components/expression/expression-renderer.js
```diff
@@ -594,7 +633,119 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+          matrixName: heatmapData.matrixname,
+          geneRegex: heatmapData.geneRegex,
+          conditionRegex: heatmapData.conditionRegex,
+          geneNames: heatmapData.geneNames,
+          conditionNames: heatmapData.conditionNames
+        });
+        this.data.zoomStack.push(currentRegex);
+        this.signal('expressionZoomIn', zoomRegex);
+      }
+      else {
```
yubowenok: pls style

src/components/expression/expression-renderer.js
yubowenok: Besides, iterating all cells for every mousemove could be slow... You may consider drawing a drag box only, and highlighting the selected rows/columns labels instead.

server/expression.js
```diff
@@ -26,22 +26,32 @@ expression.RawMatrix;
...
+ *   allGeneNames: !Array<string>,
+ *   allConditionNames: !Array<string>,
+ *   allValueMin: number,
+ *   allValueMax: number
+ * }}
+ */
+expression.MatrixAll;
+
+/**
+ * @typedef {{
```
yubowenok, line 38: Please update wiki for changed queries.
https://github.com/ViDA-NYU/genotet/wiki/Expression-Queries

server/expression.js
```diff
@@ -61,22 +78,33 @@ expression.query.Matrix;
...
 
 // Start public APIs
 /**
+ * @param {!expression.query.MatrixAll} query
+ * @param {string} expressionPath
+ * @return {?expression.MatrixAll}
+ */
+expression.query.matrixAll = function(query, expressionPath) {
+  var file = expressionPath + query.matrixName;
+  return expression.readExpressionAll_(file);
```
yubowenok: It seems that purpose is to get only rowNames, colNames and value min/max of a given matrix. The query name here is confusing, as it indicates that you want to get the entire matrix (including their values). 

Maybe you can modify the matrix query, to indicate that you don't need values to be sent back. That could save an additional query type.

server/expression.js
```diff
@@ -61,22 +78,33 @@ expression.query.Matrix;
...
+};
+
+/**
  * @param {!expression.query.Matrix} query
  * @param {string} expressionPath
  * @return {?expression.Matrix}
  */
 expression.query.matrix = function(query, expressionPath) {
```
yubowenok, line 106: Continued above: here you can include query.needValues as a boolean, and use regex selections for all rows and columns.

server/expression.js
```diff
@@ -204,72 +233,75 @@ expression.getTFAmatLine_ = function(file, name) {
...
   if (fileTFA != null && bufTFA == null) {
-    console.error('cannot read file', fileTFA);
-    return null;
+    return console.error('cannot read file', fileTFA), [];
   }
 
   var resultExp = expression.readExpmat_(bufExp);
   var resultTFA;
   if (fileTFA != null) {
-    resultTFA = expression.readTFAmat_(/** @type {!Buffer} */(bufTFA));
```
yubowenok, line 233: Removing the type coersion doesn't seem right to me.

server/expression.js
```diff
@@ -204,72 +233,75 @@ expression.getTFAmatLine_ = function(file, name) {
...
   var bufExp = utils.readFileToBuf(fileExp);
   var bufTFA = null;
   if (fileTFA != null) {
     bufTFA = utils.readFileToBuf(fileTFA);
   }
-
   if (bufExp == null) {
-    console.error('cannot read file', fileExp);
-    return null;
+    return console.error('cannot read file', fileExp), [];
```
yubowenok: Looks like you have out-of-date lines in your previous sync. You are having regression here.

server/expression.js
```diff
@@ -204,72 +233,75 @@ expression.getTFAmatLine_ = function(file, name) {
...
   }
   if (fileTFA != null && bufTFA == null) {
-    console.error('cannot read file', fileTFA);
-    return null;
+    return console.error('cannot read file', fileTFA), [];
```
yubowenok: Regression

server/expression.js
```diff
@@ -204,72 +233,75 @@ expression.getTFAmatLine_ = function(file, name) {
...
-      name = resultExp.rownames[i];
-      break;
-    }
-  }
-  if (i == resultExp.rownames.length) {
-    return null;// cannot find gene
-  }
-  var tfaValues = [];
+  if (geneNames.length == 0)
+    return []; // cannot find gene
```
yubowenok: Regression

server/expression.js
```diff
@@ -204,72 +233,75 @@ expression.getTFAmatLine_ = function(file, name) {
...
-    var tfai = resultTFA.rownames.indexOf(name);
-    if (tfai != -1) {
-      for (var j = 0; j < resultTFA.numcols; j++) {
-        var idx = resultExp.colnames.indexOf(resultTFA.colnames[j]);
-        tfaValues.push({
-          value: resultTFA.values[tfai * resultTFA.numcols + j],
-          index: idx
-        });
-      }
+    geneNames.forEach(function(geneName, i) {
```
yubowenok: Can we use a more efficient implementation better than quadratic?

server/expression.js
```diff
@@ -378,80 +410,109 @@ expression.listMatrix_ = function(expmatAddr) {
...
-    console.log('incorrect regular expression');
-    expGene = expCondition = 'a^';
-  }
   var lines = fs.readFileSync(expressionFile).toString().split('\n');
   lines.forEach(function(line) {
     var parts = line.split('\t');
     if (isFirstCol) {
-      // first column contains the conditions
+      // first row contains the conditions
       isFirstCol = false;
```
yubowenok: This should be isFirstRow.

server/expression.js
```diff
@@ -378,80 +410,109 @@ expression.listMatrix_ = function(expmatAddr) {
...
-      for (var i = 1; i < parts.length; i++) {
-        if (parts[i].match(expCondition)) {
-          conditions.push(i);
-          validCol.push(true);
-          conditionNames.push(parts[i]);
-        } else {
-          validCol.push(false);
+      conditionNames.forEach(function(conditionName) {
+        var conditionIndex = parts.indexOf(conditionName);
```
yubowenok: Most of the time, forEach and then indexOf indicate a not efficient solution.

server/expression.js
```diff
@@ -378,80 +410,109 @@ expression.listMatrix_ = function(expmatAddr) {
...
+      for (var i = 1; i < parts.length; i++) {
+        var value = parseFloat(parts[i]);
+        allValueMin = Math.min(allValueMin, value);
+        allValueMax = Math.max(allValueMax, value);
       }
     }
   });
+  Object.keys(values)
+    .sort()
+    .forEach(function(geneName, i) {
```
yubowenok:
``` js
var sortedValues = Object.keys(values)
  .sort()
  .map(function(geneName) {
    return values[geneName];
  });
```

server/expression.js
```diff
@@ -378,80 +410,109 @@ expression.listMatrix_ = function(expmatAddr) {
...
+  };
+};
+
+/**
+ * Reads all of the expression matrix data from text file.
+ * @param {string} expressionFile Path to the expression file.
+ * @return {expression.Matrix}
+ * @private
+ */
+expression.readExpressionAll_ = function(expressionFile) {
```
yubowenok: Combine this with readExpression if possible, using an extra flag. If not possible, then give it a clearer name, like getMatrixInfo.

server/server.js
```diff
@@ -210,6 +210,9 @@ app.get('/genotet', function(req, res) {
     case 'expression':
       data = expression.query.matrix(query, expressionPath);
       break;
+    case 'expression-all':
```
yubowenok: unclear and confusing query name :/

src/components/expression/expression-loader.js
```diff
@@ -24,101 +28,249 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
+      matrixName: matrixName,
+      geneNames: geneNames,
+      conditionNames: conditionNames
+    };
 
-      this.data.matrix = data;
+    $.get(genotet.data.serverURL, params, function(data) {
+        // Store the last applied data selectors.
+        _.extend(data, {
+          matrixname: matrixName,
```
yubowenok: matrixName

src/components/expression/expression-loader.js
```diff
@@ -24,101 +28,249 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
+          dataName: 'b-subtilis'
```
yubowenok: why constant here?

src/components/expression/expression-loader.js
```diff
@@ -24,101 +28,249 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
+ * @param {!Array<string>} geneNames Names for gene selection.
+ * @param {!Array<string>} conditionNames Names for experiment condition
+ *      selection.
+ * @private
+ */
+genotet.ExpressionLoader.prototype.loadTfaData_ =
+  function(matrixName, dataName, geneNames, conditionNames) {
+    var tfaParams = {
+      type: 'expression-profile',
+      matrixName: 'b-subtilis',
```
yubowenok, line 143: Should this be the current matrix name?

src/components/expression/expression-loader.js
yubowenok, line 101: Time to remove this constant.

src/components/expression/expression-loader.js
```diff
@@ -24,101 +28,249 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
+  var geneNames = [];
+  if (isGeneRegex) {
+    var geneRegex = RegExp(geneInput, 'i');
+    this.data.matrixAll.allGeneNames.forEach(function(geneName) {
+      if (geneName.match(geneRegex)) {
+        geneNames.push(geneName);
+      }
+    });
+  }
+  else {
```
yubowenok: style pls

src/components/expression/expression-loader.js
```diff
@@ -24,101 +28,249 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
-      conditionRegex += '|' + regex;
-      break;
-    case 'removeCondition':
-      // Remove the regex.
-      conditionRegex = '';
-      this.data.matrix.conditionNames.forEach(function(conditionName) {
-        if (!conditionName.toUpperCase().match(regex)) {
-          conditionRegex += conditionName + '|';
+    }
+    else {
```
yubowenok: style pls

src/components/expression/expression-loader.js
```diff
@@ -24,101 +28,249 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
+ * @param {string} conditionInput Condition input.
+ * @return {!Array<string>} conditionNames Names for experiment condition
+ *      selection.
+ */
+genotet.ExpressionLoader.prototype.formatConditionInput =
+  function(isConditionRegex, conditionInput) {
+    var conditionNames = [];
+    if (isConditionRegex) {
+      var conditionRegex = RegExp(conditionInput, 'i');
+      this.data.matrixAll.allConditionNames.forEach(function(conditionName) {
```
yubowenok:
``` js
var conditionNames = this.data.matrixAll.allConditionNames.filter(function(conditionName) {
  return conditionName.match(conditionRegex);
});
```

src/components/expression/expression-loader.js
```diff
@@ -24,101 +28,249 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
-    case 'removeCondition':
-      // Remove the regex.
-      conditionRegex = '';
-      this.data.matrix.conditionNames.forEach(function(conditionName) {
-        if (!conditionName.toUpperCase().match(regex)) {
-          conditionRegex += conditionName + '|';
+    }
+    else {
+      var inputWords = conditionInput.split(',');
+      inputWords.forEach(function(word) {
```
yubowenok: dito. use Array.filter().

src/components/expression/expression-panel.js
```diff
@@ -93,26 +97,61 @@ genotet.ExpressionPanel.prototype.initPanel = function() {
...
+      var geneNames = [];
+      if (this.isRegex_) {
+        var geneRegex = RegExp(geneInput, 'i');
+        this.data.matrixAll.allGeneNames.forEach(function(geneName) {
+          if (geneName.match(geneRegex)) {
+            geneNames.push(geneName);
+          }
+        });
+      }
+      else {
```
yubowenok: style pls

src/components/expression/expression-panel.js
```diff
@@ -122,19 +161,49 @@ genotet.ExpressionPanel.prototype.initPanel = function() {
...
+      var conditionNames = [];
+      if (this.isRegex_) {
+        var conditionRegex = RegExp(conditionInput, 'i');
+        this.data.matrixAll.allConditionNames.forEach(function(conditionName) {
+          if (conditionName.match(conditionRegex)) {
+            conditionNames.push(conditionName);
+          }
+        });
+      }
+      else {
```
yubowenok: You may want to update your IDE defaults...

src/components/expression/expression-renderer.js
```diff
@@ -141,7 +160,7 @@ genotet.ExpressionRenderer.Cell = function(params) {
  *   row: (?number|undefined),
  *   hoverColumn: (?number|undefined),
  *   hoverConditionName: (?string|undefined),
- *   hoverValue: (?number|undefined),
+ *   hoverValue: (?number|?string|undefined),
```
yubowenok: number|string|null|undefined

src/components/expression/expression-renderer.js
```diff
@@ -171,15 +190,49 @@ genotet.ExpressionRenderer.Profile = function(params) {
   this.hoverConditionName = params.hoverConditionName != null ?
     params.hoverConditionName : null;
 
-  /** @type {number} */
+  /** @type {?number|?string} */
```
yubowenok: number|string|null. avoid 2 question marks.

src/components/expression/expression-renderer.js
```diff
@@ -171,15 +190,49 @@ genotet.ExpressionRenderer.Profile = function(params) {
...
   /** @type {?string} */
   this.color = params.color != null ? params.color : null;
 };
 
+/**
+ * Zoom status object storing the status of expression matrix.
+ * @param {!{
+ *   matrixName: (?string|undefined),
+ *   dataName: (?string|undefined),
+ *   geneNames: (!Array<string>),
```
yubowenok, line 205: Brakcets are not necessary here

src/components/expression/expression-renderer.js
```diff
@@ -551,6 +713,9 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
   var heatmapRects = heatmapRows.selectAll('rect').data(_.identity);
   heatmapRects.enter().append('rect');
   heatmapRects
+    .attr('class', function(value, i, j) {
+      return 'cell row' + j + ' column' + i;
```
yubowenok: If every rect is a cell, maybe use "g.heatmap > rect" or similar, and avoid adding cell class to every rect.

src/components/expression/expression-renderer.js
```diff
@@ -594,7 +759,143 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+
+  // Zoom selection.
+  var cellWidth = this.cellWidth;
+  var cellHeight = this.cellHeight;
+  var columnStart, columnEnd, rowStart, rowEnd;
+  var zoomOutButtonSelection = d3.select('#zoom-out button');
+  if (this.data.zoomStack.length == 0) {
+    zoomOutButtonSelection.classed('disabled', true);
+  }
+  else {
```
yubowenok: style

src/components/expression/expression-renderer.js
```diff
@@ -594,7 +759,143 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
+
+        rowStart = Math.floor(selectedRange.y / cellHeight);
+        rowEnd = Math.floor((selectedRange.y + selectedRange.height) /
+          cellHeight);
+        columnStart = Math.floor(selectedRange.x / cellWidth);
+        columnEnd = Math.floor((selectedRange.x + selectedRange.width) /
+          cellWidth);
+        d3.selectAll('.gene-label')
+          .classed('label-selected', function(label, i) {
+          return i >= rowStart && i <= rowEnd;
```
yubowenok: indent 2 more spaces this line and next.

src/components/expression/expression-renderer.js
liana-yang: In heatmap content, there are not only cell rect, but also the drag selection rect. If I want to select all the cell rect, we can add cell class to every rect.

yubowenok: There can be thousands of rects and writing "cell" class to each one costs at least thousands of extra characters in DOM. Why not giving drag rect a special class, and then use a css not selector to define styles for all other rects.

src/components/expression/expression-renderer.js
liana-yang: I don't think it's necessary...

server/expression.js
liana-yang: I merged the update of refactor before. Why does here have regression?

yubowenok: You merged incorrectly. Your code is having regression. In other words, you didn't take in the refactor's new changes.

src/components/expression/expression-renderer.js
yubowenok: Up to you. You have roughly eight -1 usages related to undefined rows and columns. Making a constant is not very complex anyway.

server/expression.js
liana-yang: It is up-to-date compared to refactor before I made changes. Why regression?

yubowenok: https://github.com/ViDA-NYU/genotet/blob/refactor/server/expression.js#L243

src/components/expression/expression-renderer.js
liana-yang: This kind of comparison is very common in many files... If we give constant here, we should give constant in other files too. I don't think it is that necessary.

server/expression.js
liana-yang, line 106: Fixed.

server/expression.js
liana-yang: I cannot find any more efficient way... I think traversing the geneNames and conditionNames, then check whether they are in TFA file or find their indices are unavoidable.

src/components/expression/expression-loader.js
liana-yang, line 143: I have some doubts here. There are two names here for expression: one is "expressionMatrix" -- the expression file name (now: this.data.matrix.matrixName); another is "b-subtilis" or "rna-seq" -- the data selected from expression create UI (now: this.data.matrix.dataName). Now dataName did nothing in expression query but you said it is related to expression data too. I am confused about the roles these two names play in the expression data.

liana-yang, line 143: However, the expression create UI only have dataName selection. Where is the matrixName selection?

src/components/expression/expression-renderer.js
liana-yang: What if I want to selectAll('.cell')?

yubowenok: can't you do selectAll('rect:not(...)')?

src/components/expression/expression-loader.js
yubowenok, line 143: b-subtilis and rna-seq are matrixNames. matrixNames are the same as dataNames. Maybe use matrixName uniformly.

yubowenok, line 143: Besides the server side will handle fileName as primary key for data selection. So you may want to adjust accordingly.

server/expression.js
yubowenok: Can't you just create some search data structures for the indices?

``` js
geneNames.forEach(function(geneName) {
  var tfai = tfaRowIndices[geneName]; // <-- create a DS for your indices
  // ...
}
```

src/components/expression/expression-renderer.js
yubowenok: You are using a special meaning of -1 that indicates undefined row or column. It is not good to have many such constants here and there, unless the constant's meaning is well-known. If you are using indexOf() == -1, that's fine. However, your -1 has special indications and could be better handled by "enum"-like constants instead.

I'm not aware of other files using -1 for non indexOf operations. If they are indeed using, they need to either detailedly comment the usage, or define a constant as well.

server/expression.js
```diff
@@ -204,20 +233,21 @@ expression.getTFAmatLine_ = function(file, name) {
...
  * @param {string} fileExp File name of the expression matrix.
  * @param {string} fileTFA File name of the TFA matrix.
- * @param {string} name Name of the gene to be profiled.
+ * @param {!Array<string>} geneNames Names of the expression matrix.
+ * @param {!Array<string>} conditionNames Names of the expression matrix.
  * @return {?expression.Profile} Gene expression profile as a JS object.
  * @private
  */
-expression.getExpmatLine_ = function(fileExp, fileTFA, name) {
+expression.getExpmatLine_ = function(fileExp, fileTFA, geneNames,
```
yubowenok: Maybe getProfile_ is a better name. Please replace the obsolete method name.

dev/view.js
```diff
@@ -13,8 +13,11 @@ genotet.test = function() {
 
   genotet.viewManager.createView('expression', 'My Expression Matrix', {
     matrixName: 'expressionMatrix',
-    geneRegex: 'sig.*',
-    condRegex: 'SigH.*'
+    dataName: 'B. Subtilis',
```
yubowenok: I don't think dataName is defined? matrixName corresponds to the data identifier.

src/components/expression/expression-renderer.js
```diff
@@ -487,7 +623,23 @@ genotet.ExpressionRenderer.prototype.dataLoaded = function() {
       i++;
     }
   }
+
+  i = 0;
+  profileCount = this.data.tfaProfiles.length;
+  while (i < profileCount) {
+    var geneIndex = heatmapData.geneNames.indexOf(
```
yubowenok: You are querying the geneNames indices in two passes (another pass above for profiles) using this non-efficient indexOf. You can create a DS that stores the indices for heatmap.geneNames and perform the index retrieval much more efficiently.

src/components/expression/expression-loader.js
liana-yang, line 143: I will change this part after @djasmine finish the matrixName and fileName separation part.

server/expression.js
```diff
@@ -233,43 +263,55 @@ expression.getExpmatLine_ = function(fileExp, fileTFA, name) {
...
-    if (resultExp.rownames[i].toLowerCase() == name) {
-      name = resultExp.rownames[i];
-      break;
-    }
-  }
-  if (i == resultExp.rownames.length) {
-    return null;// cannot find gene
-  }
-  var tfaValues = [];
+  if (geneNames.length == 0)
```
yubowenok: brackets pls

server/expression.js
```diff
@@ -233,43 +263,55 @@ expression.getExpmatLine_ = function(fileExp, fileTFA, name) {
...
+    });
+    resultTFA.colnames.forEach(function(conditionName, i) {
+      allConditionNames[conditionName] = i;
+    });
+    geneNames.forEach(function(geneName, i) {
+      var tfai = allGeneNames[geneName];
+      var tfaValues = [];
+      conditionNames.forEach(function(conditionName, i) {
+        var idx = allConditionNames[conditionName];
+        if (typeof idx != 'undefined') {
```
yubowenok:
``` js
if (conditionName in allConditionNames) {
  var idx = allConditionNames[conditionName];
  // ...
}
```

server/expression.js
```diff
@@ -233,43 +263,55 @@ expression.getExpmatLine_ = function(fileExp, fileTFA, name) {
...
+    resultTFA.rownames.forEach(function(geneName, i) {
+      allGeneNames[geneName] = i;
+    });
+    resultTFA.colnames.forEach(function(conditionName, i) {
+      allConditionNames[conditionName] = i;
+    });
+    geneNames.forEach(function(geneName, i) {
+      var tfai = allGeneNames[geneName];
+      var tfaValues = [];
+      conditionNames.forEach(function(conditionName, i) {
```
yubowenok: Outside variable "i" is shadowed and ambiguous.

server/expression.js
```diff
@@ -233,43 +263,55 @@ expression.getExpmatLine_ = function(fileExp, fileTFA, name) {
...
+        var idx = allConditionNames[conditionName];
+        if (typeof idx != 'undefined') {
+          var tfaValue = resultTFA.values[tfai * resultTFA.numcols + i];
+          if (tfaValue) {
+            tfaValues.push({
+              value: tfaValue,
+              index: i
```
yubowenok: Are you taking the inner "i"? Please rename to avoid shadowed variables.

server/expression.js
```diff
@@ -378,72 +420,53 @@ expression.listMatrix_ = function(expmatAddr) {
...
-      }
+      });
+    } else {
+      // other rows contain a gene, and values
+      allGeneNames[parts[0]] = lineIndex;
+    }
+  });
+  geneNames.forEach(function(geneName) {
+    var geneIndex = allGeneNames[geneName];
+    if (typeof geneIndex != 'undefined') {
```
yubowenok: ditto

src/components/expression/expression-loader.js
```diff
@@ -24,101 +31,210 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
+        names.forEach(function(name) {
+          zoomStatus.conditionNames.push(name);
+        });
+        break;
+      case 'removeCondition':
+        // Remove the names.
+        this.removeNames_(zoomStatus.conditionNames, names);
+        break;
+    }
+  }.bind(this));
```
yubowenok:
``` js
...forEach(function(){}, this);
```

The 2nd param of forEach is "this" object. No need to bind.

src/components/expression/expression-loader.js
```diff
@@ -24,101 +31,210 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
+ * Remove genes or conditions from previous status in expression.
+ * @param {!Array<string>} originalNames Original names from previous zoom
+ *      status.
+ * @param {!Array<string>} removeNames Names need to be removed.
+ * @private
+ */
+genotet.ExpressionLoader.prototype.removeNames_ = function(originalNames,
+                                                           removeNames) {
+  removeNames.forEach(function(name) {
+    var geneIndex = originalNames.indexOf(name);
```
yubowenok: Use DS here too.

src/components/expression/expression-panel.js
```diff
@@ -93,26 +97,45 @@ genotet.ExpressionPanel.prototype.initPanel = function() {
   this.selectProfiles_
     .on('select2:select', function(event) {
       var geneIndex = event.params.data.element.index;
+      var geneName = event.params.data.text;
       this.signal('addGeneProfile', geneIndex);
+      this.signal('addTfaProfile', geneName);
```
yubowenok: If these two events are always signaled together, why not just using one?

src/components/expression/expression-panel.js
```diff
@@ -219,3 +257,51 @@ genotet.ExpressionPanel.prototype.displayCellInfo = function(geneName,
...
+  if (isGeneRegex) {
+    var geneRegex = RegExp(geneInput, 'i');
+    geneNames = Object.keys(this.data.matrixInfo.allGeneNames)
+      .filter(function(geneName) {
+        return geneName.match(geneRegex);
+      });
+  } else {
+    var inputWords = geneInput.split(',');
+    geneNames = inputWords.filter(function(word) {
+      return typeof this.data.matrixInfo.allGeneNames[word] != 'undefined';
```
yubowenok: Please use "in".

src/components/expression/expression-panel.js
```diff
@@ -219,3 +257,51 @@ genotet.ExpressionPanel.prototype.displayCellInfo = function(geneName,
...
+    if (isConditionRegex) {
+      var conditionRegex = RegExp(conditionInput, 'i');
+      conditionNames = Object.keys(this.data.matrixInfo.allConditionNames)
+        .filter(function(conditionName) {
+          return conditionName.match(conditionRegex);
+        });
+    } else {
+      var inputWords = conditionInput.split(',');
+      conditionNames = inputWords.filter(function(word) {
+        return typeof this.data.matrixInfo.allConditionNames[word] !=
```
yubowenok: ditto

src/components/expression/expression-renderer.js
```diff
@@ -477,9 +613,9 @@ genotet.ExpressionRenderer.prototype.dataLoaded = function() {
   var i = 0;
   var profileCount = this.data.profiles.length;
   while (i < profileCount) {
-    var geneIndex = heatmapData.geneNames.indexOf(
-      this.data.profiles[i].geneName);
-    if (geneIndex == -1) {
+    var geneIndex = this.data.matrixGeneNameDict[
+      this.data.profiles[i].geneName];
+    if (typeof geneIndex == 'undefined') {
```
yubowenok: ditto

src/components/expression/expression-renderer.js
```diff
@@ -487,7 +623,23 @@ genotet.ExpressionRenderer.prototype.dataLoaded = function() {
       i++;
     }
   }
+
+  i = 0;
+  profileCount = this.data.tfaProfiles.length;
+  while (i < profileCount) {
+    var geneIndex = this.data.matrixGeneNameDict[
+      this.data.tfaProfiles[i].geneName];
+    if (typeof geneIndex == 'undefined') {
```
yubowenok: ditto

src/components/expression/expression-renderer.js
```diff
@@ -594,7 +758,141 @@ genotet.ExpressionRenderer.prototype.drawMatrixCells_ = function() {
...
   heatmapRects.exit().remove();
-  this.highlightLabelsAfterUpdateData_();
+
+  // Zoom selection.
+  var cellWidth = this.cellWidth;
+  var cellHeight = this.cellHeight;
+  var columnStart, columnEnd, rowStart, rowEnd;
+  var zoomOutButtonSelection = d3.select('#zoom-out button');
+  if (this.data.zoomStack.length == 0) {
+    zoomOutButtonSelection.classed('disabled', true);
```
``` js
zoomOutButtonSelection.classed('disabled', this.data.zoomStack.length == 0);
```

src/components/expression/expression-loader.js
liana-yang: I think it is unnecessary. Because originalNames is an array and the time complexity of this line is O(n). However, the time complexity of building a dict for an array is also O(n). Therefore, I think it is not that necessary to use DS here.

yubowenok: Apparently, forEach of removeNames is also linear. This makes these two lines quadratic.

server/expression.js
```diff
@@ -233,43 +266,57 @@ expression.getExpmatLine_ = function(fileExp, fileTFA, name) {
...
+      allTfaValues.push(tfaValues);
+    });
   }
-  var values = [];
-  for (var j = 0; j < resultExp.numcols; j++) {
-    values.push(resultExp.values[i * resultExp.numcols + j]);
-  }
-  console.log('returning line', name);
+  var geneNameString = geneNames.join(',');
+  console.log('returning tfa line', geneNameString);
```
yubowenok:
``` js
console.log('returning tfa line', geneNames.join(','));
```

src/components/expression/expression-loader.js
```diff
@@ -24,101 +31,213 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
-    matrixName: matrixName,
-    geneRegex: geneRegex,
-    conditionRegex: conditionRegex
+genotet.ExpressionLoader.prototype.loadExpressionMatrixInfo =
+  function(matrixName, dataName) {
+    var params = {
+      type: 'expression-info',
+      matrixName: matrixName
+    };
+
```
yubowenok, line 57: You didn't fire a loadStart/loadComplete here. Merge #55 and use the wrapper.

src/components/expression/expression-loader.js
```diff
@@ -24,101 +31,213 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
+        if (data.conditionNames.length == 0) {
+          genotet.warning('input condition not found');
+          return;
+        }
 
-      this.signal('loadComplete');
-    }.bind(this), 'jsonp')
-    .fail(this.fail.bind(this, 'cannot load expression matrix', params));
-};
+        this.signal('loadStart');
```
yubowenok: Incorrect placement of loadStart. You need to place loadStart before ajax call. Merge #55 and use wrapper.

src/components/expression/expression-loader.js
```diff
@@ -24,101 +31,213 @@ genotet.utils.inherit(genotet.ExpressionLoader, genotet.ViewLoader);
...
+          return;
+        }
+        var tfaGeneNameDict = {};
+        data.geneNames.forEach(function(geneName, i) {
+          tfaGeneNameDict[geneName] = i;
+        }.bind(this));
+        this.data.tfaGeneNameDict = tfaGeneNameDict;
+        this.data.tfaData = data;
+
+        this.signal('loadComplete');
```
yubowenok: loadComplete doesn't seem to match any loadStart.

src/components/expression/expression-loader.js
liana-yang, line 57: Fixed.

yubowenok:
LGTM

PR #54, opened 2016-01-06
type checking for server and testing code

yubowenok:
- type checking for server and testing code
- refine gulp tasks

Type checking heavily uses [externs](https://developers.google.com/closure/compiler/docs/api-tutorial3?hl=en#externs). Please read and try to understand the behavior of externs. We both use existing externs from public repos as well as our own (as some externs are a bit old and out-of-date). You may modify extern files when necessary. However please try to modify as little as possible the externs and only change extern in a reasonable way.

Github doesn't support multiple assignees @djasmine @Shauro. Please both confirm before I merge.


djasmine:
confirmed

liana-yang:
confirmed.

yubowenok:
More info [here](https://github.com/ViDA-NYU/genotet/wiki/Server-Code-Compilation).

PR #71, opened 2016-01-20
Fixed bed track position and binding issues.

liana-yang:


src/components/base/view.js
```diff
@@ -119,6 +119,9 @@ genotet.View.prototype.init = function() {
     .on('genotet.loadFail', function() {
       this.renderer.showFailure();
     }.bind(this));
+
+  // Override jQuery Resizable handle.
+  $('.ui-resizable-handle').css('z-index', 'auto');
```
yubowenok: You are setting z-index for resizable-handles globally. However you should set z-index for handles only in this view.

Note that this method works only when the resizable-handles have a stable DOM order in the view. This is the case for now.

yubowenok: Have you fixed this @Shauro ?

yubowenok:
LGTM

src/components/binding/binding-renderer.js
```diff
@@ -124,9 +128,11 @@ genotet.BindingRenderer.prototype.EXON_LABEL_SIZE = 6;
 genotet.BindingRenderer.prototype.BED_LABEL_SIZE = 13;
 /** @const {number} */
 genotet.BindingRenderer.prototype.BED_MIN_WIDTH = 3;
+/** @const {number} */
+genotet.BindingRenderer.prototype.BED_HEIGHT_PROPORTION = 0.8;
 
 /** @const {!Array<number>} */
-genotet.BindingRenderer.ZOOM_EXTENT = [1, 65536];
+genotet.BindingRenderer.ZOOM_EXTENT = [1, 15673503];
```
yubowenok: Why 15673503?
http://www.wolframalpha.com/input/?i=15673503

liana-yang: I calculated it in this way: 
the range of overview: range = 197194699 - 2999997 = 194194702,
the width of canvas: width = 1239px,
the width of bed label (majority amount): label = 100px,
to guarantee that the labels of neighbor bed rects could be seen clearly, we need the max zoom extent: maxExtent = range \* label / width = 15673503.

yubowenok: This doesn't make sense at all. The range of overview depends on data.

PR #77, opened 2016-01-21
Fixed expression issues.

liana-yang:
#56 Fixed selection shifting on heatmap box zoom in.
#57 Ignored cases when input string in expression.
#63 Added legend to expression TFA profile.
#68 Added one more input style checkbox in expression panel and fix their sizes.

### DIFF
--- a/src/components/expression/expression-loader.js
+++ b/src/components/expression/expression-loader.js
@@ -19,6 +19,8 @@ genotet.ExpressionLoader = function(data) {
     tfaData: null,
     matrixGeneNameDict: null,
     matrixConditionNameDict: null,
+    lowerGeneNames: null,
+    lowerConditionNames: null,
     tfaGeneNameDict: null,
     profiles: [],
     tfaProfiles: [],
@@ -106,6 +108,7 @@ genotet.ExpressionLoader.prototype.loadExpressionMatrix_ = function(fileName,
     }
 
     this.data.matrix = data;
+
     var matrixGeneNameDict = {};
     data.geneNames.forEach(function(geneName, i) {
       matrixGeneNameDict[geneName] = i;
@@ -116,6 +119,18 @@ genotet.ExpressionLoader.prototype.loadExpressionMatrix_ = function(fileName,
     }.bind(this));
     this.data.matrixGeneNameDict = matrixGeneNameDict;
     this.data.matrixConditionNameDict = matrixConditionNameDict;
+
+    var lowerGeneNames = {};
+    Object.keys(this.data.matrixInfo.allGeneNames).forEach(function(name) {
+      lowerGeneNames[name.toLowerCase()] = name;
+    });
+    this.data.lowerGeneNames = lowerGeneNames;
+    var lowerConditionNames = {};
+    Object.keys(this.data.matrixInfo.allConditionNames).forEach(
+      function(name) {
+        lowerConditionNames[name.toLowerCase()] = name;
+      });
+    this.data.lowerConditionNames = lowerConditionNames;
 
     this.loadTfaProfile_(fileName, geneNames, conditionNames);
   }.bind(this), 'cannot load expression matrix');
@@ -212,26 +227,12 @@ genotet.ExpressionLoader.prototype.update = function(method, fileName,
  * @param {!Array<string>} originalNames Original names from previous zoom
  *      status.
  * @param {!Array<string>} removeNames Names need to be removed.
- * @return {!Array<string>} result Revised names.
+ * @return {!Array<string>}
  * @private
  */
 genotet.ExpressionLoader.prototype.removeNames_ = function(originalNames,
                                                            removeNames) {
-  var namesDict = {};
-  var result = [];
-  originalNames.forEach(function(name, i) {
-    namesDict[name] = i;
+  return originalNames.filter(function(name) {
+    return removeNames.indexOf(name) < 0;
   });
-  removeNames.forEach(function(name) {
-    var geneIndex = namesDict[name];
-    if (name in namesDict) {
-      delete originalNames[geneIndex];
-    }
-  });
-  originalNames.forEach(function(name) {
-    if (name) {
-      result.push(name);
-    }
-  });
-  return result;
 };
--- a/src/components/expression/expression-loader.js
+++ b/src/components/expression/expression-loader.js
@@ -232,7 +232,11 @@ genotet.ExpressionLoader.prototype.update = function(method, fileName,
  */
 genotet.ExpressionLoader.prototype.removeNames_ = function(originalNames,
                                                            removeNames) {
+  var removeNamesDict = {};
+  removeNames.forEach(function(name, i) {
+    removeNamesDict[name] = i;
+  });
   return originalNames.filter(function(name) {
-    return removeNames.indexOf(name) < 0;
+    return !(name in removeNamesDict);
   });
 };
--- a/src/components/expression/expression-loader.js
+++ b/src/components/expression/expression-loader.js
@@ -162,7 +162,6 @@ genotet.ExpressionLoader.prototype.update = function(method, fileName,
                                                      names) {
   var heatmapData = this.data.matrix;
   var currentStatus = new genotet.ExpressionRenderer.ZoomStatus({
-    fileName: heatmapData.fileName,
     geneNames: heatmapData.geneNames,
     conditionNames: heatmapData.conditionNames
   });
@@ -175,14 +174,14 @@ genotet.ExpressionLoader.prototype.update = function(method, fileName,
         break;
       case 'addGene':
         // Concat the two names.
-        this.removeNames_(zoomStatus.geneNames, names);
+        zoomStatus.geneNames = this.removeNames_(zoomStatus.geneNames, names);
         names.forEach(function(name) {
           zoomStatus.geneNames.push(name);
         });
         break;
       case 'removeGene':
         // Remove the names.
-        this.removeNames_(zoomStatus.geneNames, names);
+        zoomStatus.geneNames = this.removeNames_(zoomStatus.geneNames, names);
         break;
       case 'setCondition':
         // Totally replace the regex.
@@ -190,19 +189,21 @@ genotet.ExpressionLoader.prototype.update = function(method, fileName,
         break;
       case 'addCondition':
         // Concat the two names.
-        this.removeNames_(zoomStatus.conditionNames, names);
+        zoomStatus.conditionNames = this.removeNames_(zoomStatus.conditionNames,
+          names);
         names.forEach(function(name) {
           zoomStatus.conditionNames.push(name);
         });
         break;
       case 'removeCondition':
         // Remove the names.
-        this.removeNames_(zoomStatus.conditionNames, names);
+        zoomStatus.conditionNames = this.removeNames_(zoomStatus.conditionNames,
+          names);
         break;
     }
   }, this);
   var zoomStatus = this.data.zoomStack.pop();
-  this.load(zoomStatus.fileName, zoomStatus.geneNames,
+  this.load(heatmapData.fileName, zoomStatus.geneNames,
     zoomStatus.conditionNames);
 };
 
@@ -211,18 +212,26 @@ genotet.ExpressionLoader.prototype.update = function(method, fileName,
  * @param {!Array<string>} originalNames Original names from previous zoom
  *      status.
  * @param {!Array<string>} removeNames Names need to be removed.
+ * @return {!Array<string>} result Revised names.
  * @private
  */
 genotet.ExpressionLoader.prototype.removeNames_ = function(originalNames,
                                                            removeNames) {
   var namesDict = {};
+  var result = [];
   originalNames.forEach(function(name, i) {
     namesDict[name] = i;
   });
   removeNames.forEach(function(name) {
     var geneIndex = namesDict[name];
     if (name in namesDict) {
-      originalNames.splice(geneIndex, 1);
+      delete originalNames[geneIndex];
     }
   });
+  originalNames.forEach(function(name) {
+    if (name) {
+      result.push(name);
+    }
+  });
+  return result;
 };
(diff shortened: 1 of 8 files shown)


src/components/expression/expression-loader.js
```diff
@@ -211,18 +212,26 @@ genotet.ExpressionLoader.prototype.update = function(method, fileName,
...
   var namesDict = {};
+  var result = [];
   originalNames.forEach(function(name, i) {
     namesDict[name] = i;
   });
   removeNames.forEach(function(name) {
     var geneIndex = namesDict[name];
     if (name in namesDict) {
-      originalNames.splice(geneIndex, 1);
+      delete originalNames[geneIndex];
```
yubowenok: You should not delete array element using index. If you want to unset it, use explicit value assignment.

src/components/expression/expression-loader.js
```diff
@@ -211,18 +212,26 @@ genotet.ExpressionLoader.prototype.update = function(method, fileName,
...
     }
   });
+  originalNames.forEach(function(name) {
```
yubowenok: Use filter()

src/components/expression/expression-panel.html
```diff
@@ -8,18 +8,15 @@
...
-      <label id="string" class="radio-inline"><input type="radio" name="optradio">string</label>
-    </div>
-  </div>
   <div id="genes" class="panel-section">
     <div class="item-title">Genes</div>
     <!--<button class="btn btn-default glyphicon glyphicon-refresh btn-update"></button>-->
     <!--<input class="form-control input-with-update">-->
     <input class="form-control">
+    <div class="input-checkbox">
+      <label id="gene-regex" class="radio-inline"><input type="radio" name="gene-optradio">regex</label>
```
yubowenok: Too many ids in sub-elements. You may want to move is to parent container.

src/components/expression/expression-loader.js
```diff
@@ -211,18 +212,26 @@ genotet.ExpressionLoader.prototype.update = function(method, fileName,
  * @param {!Array<string>} originalNames Original names from previous zoom
  *      status.
  * @param {!Array<string>} removeNames Names need to be removed.
+ * @return {!Array<string>} result Revised names.
```
yubowenok: Unclear description of return values. Why removing names would return a list? You need to explain in detail in comments.

src/components/expression/expression-panel.html
```diff
@@ -31,6 +28,10 @@
     <!--<button class="btn btn-default glyphicon glyphicon-refresh btn-update"></button>-->
     <!--<input class="form-control input-with-update">-->
     <input class="form-control">
+    <div class="input-checkbox">
+      <label id="condition-regex" class="radio-inline"><input type="radio" name="condition-optradio">regex</label>
+      <label id="condition-string" class="radio-inline"><input type="radio" name="condition-optradio">string</label>
```
yubowenok: Ditto. Move id to parent container

src/components/expression/expression-panel.js
```diff
@@ -272,8 +285,16 @@ genotet.ExpressionPanel.prototype.formatGeneInput = function(isGeneRegex,
...
   } else {
     var inputWords = geneInput.split(',');
-    geneNames = inputWords.filter(function(word) {
-      return word in this.data.matrixInfo.allGeneNames;
+    var lowerGeneNames = {};
+    Object.keys(this.data.matrixInfo.allGeneNames).forEach(function(name) {
+      lowerGeneNames[name.toLowerCase()] = name;
+    });
+    geneNames = [];
+    inputWords.forEach(function(word) {
```
yubowenok: Use filter().

src/components/expression/expression-panel.js
```diff
@@ -296,8 +317,17 @@ genotet.ExpressionPanel.prototype.formatConditionInput =
...
       var inputWords = conditionInput.split(',');
-      conditionNames = inputWords.filter(function(word) {
-        return word in this.data.matrixInfo.allConditionNames;
+      var lowerConditionNames = {};
+      Object.keys(this.data.matrixInfo.allConditionNames).forEach(
+        function(name) {
+          lowerConditionNames[name.toLowerCase()] = name;
+        });
+      conditionNames = [];
+      inputWords.forEach(function(word) {
```
yubowenok: Use filter().

src/components/expression/expression-renderer.js
```diff
@@ -1231,6 +1209,56 @@ genotet.ExpressionRenderer.prototype.drawTfaProfiles_ = function() {
 };
 
 /**
+ * Renders legend for gene profiles and TFA profiles.
+ * @param {!Array<genotet.ExpressionRenderer.Profile>} profiles
```
yubowenok, line 1215: Are you rendering legends twice if both profile and tfa are shown?

src/components/expression/expression-renderer.js
```diff
@@ -1140,7 +1113,12 @@ genotet.ExpressionRenderer.prototype.drawTfaProfiles_ = function() {
     this.profileMargins_.left,
     this.canvasWidth - this.profileMargins_.right
   ]).domain([0, heatmapData.conditionNames.length]);
-  var yScaleTop = this.profileMargins_.top;
+  var yScaleTop = this.profileMargins_.top +
+    this.DEFAULT_PROFILE_LEGEND_HEIGHT;
+  if (this.data.tfaProfiles.length == 0 ||
```
yubowenok: !length is preferred over length==0

src/components/expression/expression-panel.js
```diff
@@ -272,8 +285,16 @@ genotet.ExpressionPanel.prototype.formatGeneInput = function(isGeneRegex,
       });
   } else {
     var inputWords = geneInput.split(',');
-    geneNames = inputWords.filter(function(word) {
-      return word in this.data.matrixInfo.allGeneNames;
+    var lowerGeneNames = {};
+    Object.keys(this.data.matrixInfo.allGeneNames).forEach(function(name) {
```
yubowenok: This isn't a good idea. You should process your lowercase dictionary only once, instead of every time user input changes.

src/components/expression/expression-panel.js
liana-yang: I think using forEach() is better. Because we want to return an array of geneNames, which are case-sensitive. filter() is used to pick up items from inputWords, which are case-insensitive.

src/components/expression/expression-panel.js
liana-yang: ditto.

src/components/expression/expression-renderer.js
liana-yang, line 1215: Fixed.

src/components/expression/expression-panel.js
```diff
@@ -272,8 +283,11 @@ genotet.ExpressionPanel.prototype.formatGeneInput = function(isGeneRegex,
       });
   } else {
     var inputWords = geneInput.split(',');
-    geneNames = inputWords.filter(function(word) {
-      return word in this.data.matrixInfo.allGeneNames;
+    inputWords.forEach(function(word) {
```
yubowenok: Use forEach's 2nd argument for "this".

templates/create-expression.html
```diff
@@ -26,21 +26,21 @@ <h4 class="modal-title">Create Expression Matrix</h4>
...
     <tr>
       <td>Genes</td>
-      <td><input id="gene-regex" class="form-control" size="20" title="regex for genes to be shown in the heatmap"></td>
+      <td><input id="gene-input" class="form-control" size="20" title="input for genes to be shown in the heatmap"></td>
       <td>
-        <div class="input-checkbox">
-          <label id="gene-input-regex" class="radio-inline"><input type="radio" name="gene-optradio">regex</label>
-          <label id="gene-input-string" class="radio-inline"><input type="radio" name="gene-optradio">string</label>
+        <div id="gene-input-type" class="input-checkbox">
+          <label class="radio-inline regex"><input type="radio" name="gene-optradio">regex</label>
```
yubowenok: Please use name or role.

src/components/expression/expression-loader.js
```diff
@@ -211,18 +227,12 @@ genotet.ExpressionLoader.prototype.update = function(method, fileName,
...
-  originalNames.forEach(function(name, i) {
-    namesDict[name] = i;
-  });
-  removeNames.forEach(function(name) {
-    var geneIndex = namesDict[name];
-    if (name in namesDict) {
-      originalNames.splice(geneIndex, 1);
-    }
+  return originalNames.filter(function(name) {
+    return removeNames.indexOf(name) < 0;
```
yubowenok: This is inefficient. You still need a dict to use filter().

src/components/expression/expression-panel.html
```diff
@@ -8,18 +8,15 @@
...
-      <label id="string" class="radio-inline"><input type="radio" name="optradio">string</label>
-    </div>
-  </div>
   <div id="genes" class="panel-section">
     <div class="item-title">Genes</div>
     <!--<button class="btn btn-default glyphicon glyphicon-refresh btn-update"></button>-->
     <!--<input class="form-control input-with-update">-->
     <input class="form-control">
+    <div id="gene-input" class="input-checkbox">
+      <label class="radio-inline" name="regex"><input type="radio" name="gene-optradio" checked>regex</label>
```
yubowenok: redundant <input> names

src/components/expression/expression-panel.html
```diff
@@ -31,6 +28,10 @@
     <!--<button class="btn btn-default glyphicon glyphicon-refresh btn-update"></button>-->
     <!--<input class="form-control input-with-update">-->
     <input class="form-control">
+    <div id="condition-input" class="input-checkbox">
+      <label class="radio-inline" name="regex"><input type="radio" name="condition-optradio" checked>regex</label>
```
yubowenok: redundant <input> names

src/components/expression/expression-panel.js
```diff
@@ -107,13 +113,22 @@ genotet.ExpressionPanel.prototype.initPanel = function() {
...
 
   // Input type update
-  this.container.find('#regex input').click(function() {
-    this.isRegex_ = true;
-  }.bind(this));
-  this.container.find('#string input').click(function() {
-    this.isRegex_ = false;
-  }.bind(this));
-  this.container.find('#regex input').trigger('click');
+  this.container.find('#gene-input label[name="regex"] input')
```
yubowenok: Why not putting names on input and do "#gene-input input[name=regex]"?

src/components/expression/expression-panel.js
```diff
@@ -272,9 +287,12 @@ genotet.ExpressionPanel.prototype.formatGeneInput = function(isGeneRegex,
       });
   } else {
     var inputWords = geneInput.split(',');
```
yubowenok: Please handle spaces for comma separated strings.

src/dialog.js
```diff
@@ -235,21 +235,19 @@ genotet.dialog.createExpression_ = function() {
         });
 
       // Choose input type of gene and condition.
-      modal.find('#gene-input-regex input')
+      modal.find('#gene-input-type label[name="regex"] input')
```
yubowenok: ditto

templates/create-expression.html
```diff
@@ -26,21 +26,21 @@ <h4 class="modal-title">Create Expression Matrix</h4>
...
     <tr>
       <td>Genes</td>
-      <td><input id="gene-regex" class="form-control" size="20" title="regex for genes to be shown in the heatmap"></td>
+      <td><input id="gene-input" class="form-control" size="20" title="input for genes to be shown in the heatmap"></td>
       <td>
-        <div class="input-checkbox">
-          <label id="gene-input-regex" class="radio-inline"><input type="radio" name="gene-optradio">regex</label>
-          <label id="gene-input-string" class="radio-inline"><input type="radio" name="gene-optradio">string</label>
+        <div id="gene-input-type" class="input-checkbox">
+          <label class="radio-inline" name="regex"><input type="radio" name="gene-optradio" checked>regex</label>
```
yubowenok: Let's set default to comma separated strings and change the order to string/regex. BTW string may be a word too CS. Is there a better choice? Maybe "text"?

src/components/expression/expression-panel.js
liana-yang: input:radio elements use the same [name] so that the browser knows to only allow one to be :checked at a time.

yubowenok: Oh I see. Please ignore comments regarding input names. 

On the other hand are the double quotes redudant? I think you can do [name=regex].

liana-yang: They are necessary: https://api.jquery.com/attribute-equals-selector/

yubowenok: Your link just says in bold text:
value: An attribute value. Can be either an unquoted single word or a quoted string.

Why necessary? :)

liana-yang: Oh... That's my fault... ＝。＝

src/components/expression/expression-panel.js
```diff
@@ -271,10 +286,15 @@ genotet.ExpressionPanel.prototype.formatGeneInput = function(isGeneRegex,
         return geneName.match(geneRegex);
       });
   } else {
+    geneInput = geneInput.replace(/^[,\s]+|[,\s]+$/g, '');
+    geneInput = geneInput.replace(/\s*,\s*/g, ',');
```
yubowenok: You can get all the tokens in one step using a single split or match.

src/components/expression/expression-panel.js
```diff
@@ -271,10 +286,15 @@ genotet.ExpressionPanel.prototype.formatGeneInput = function(isGeneRegex,
...
     var inputWords = geneInput.split(',');
-    geneNames = inputWords.filter(function(word) {
-      return word in this.data.matrixInfo.allGeneNames;
-    }.bind(this));
+    inputWords.forEach(function(word) {
```
``` js
inputWords = inputWords.toLowerCase().split(...);
geneNames = inputWords.reduce(function(result, name) {
  if (name in this.data.lowerGeneNames) {
    result.push(this.data.lowerCaseNames[name]);
  }
  return result;
}.bind(this), [])
```

yubowenok:
LGTM

PR #85, opened 2016-02-01
add mapping server handler

djasmine:
add mapping files server listing function
add mapping files server query
handle mapping files upload in front-end and server


yubowenok:
I have to run now. Will check tmrw morning.

src/css/dialog.css
```diff
@@ -0,0 +1,41 @@
+.modal-body td {
```
yubowenok: This file shouldn't be committed. You need to double-check your git usage to avoid similar things happening again.

yubowenok:
Could you please add wiki for the query?

server/mapping.js
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Lists all the mapping files.
+ * @param {string} mappingPath Path to the mapping file folder.
+ * @return {{
+ *   fileList: ?Array<string>
+ * }}
+ */
+mapping.query.list = function(mappingPath) {
+  return {
```
yubowenok: Why not returning the fileList object itself?

server/mapping.js
```diff
@@ -0,0 +1,66 @@
...
+};
+
+/**
+ * Gets mapping rules.
+ * @param {string} filePath Path to the mapping file.
+ * @return {{mappingName: !Object<string>}}
+ * @private
+ */
+mapping.getMapping_ = function(filePath) {
+  var mappingName = {};
```
yubowenok: What the meaning of "mappingName"? I think "mapping" or "mappingRules" would be better. Besides you can return the object ifself.

src/dialog.js
```diff
@@ -286,6 +286,13 @@ genotet.dialog.upload_ = function() {
       var btnUpload = modal.find('#btn-upload').prop('disabled', true);
       var btnFile = modal.find('#btn-file');
       var fileDisplay = modal.find('#file-display');
+
+      selectpicker.change(function() {
+        var isMapping = selectpicker.val() == 'mapping';
+        modal.find('#data-name').prop('disabled', isMapping);
```
yubowenok: You can hide the inputs instead of disable them.

yubowenok:
Who is assigned to complete the UI that selects the mapping file? @djasmine @Shauro

djasmine:
I think @Shauro will do the UI for selecting the mapping file. We will check it out.

yubowenok:
Have you updated the wiki for mapping queries? @djasmine

server/server.js
```diff
@@ -147,6 +156,9 @@ app.post('/genotet/upload', upload.single('file'), function(req, res) {
     case 'bed':
       prefix = bedPath;
       break;
+    case 'mapping':
```
yubowenok, line 159: Please add in readme.md to instruct user to add mappingPath in config.

src/dialog.js
```diff
@@ -291,6 +291,19 @@ genotet.dialog.upload_ = function() {
       var btnUpload = modal.find('#btn-upload').prop('disabled', true);
       var btnFile = modal.find('#btn-file');
       var fileDisplay = modal.find('#file-display');
+
+      selectpicker.change(function() {
+        var isMapping = selectpicker.val() == 'mapping';
+        if (isMapping) {
+          console.log(modal.find('#data-name'));
+          modal.find('#data-name-row').prop('style', 'display:none');
```
yubowenok:
``` js
modal.find('#data-name-row').css('display', 'none');
```

Same for the below.

templates/upload.html
```diff
@@ -16,18 +16,19 @@ <h4 class="modal-title">Upload Data</h4>
           <option value="binding">Binding Track</option>
           <option value="bed">Bed Track</option>
           <option value="expression">Expression Matrix</option>
+          <option value="mapping">Mapping</option>
         </select>
       </td>
     </tr>
-    <tr>
+    <tr id="data-name-row">
```
yubowenok: This is a bit too many ids. You can do 

``` js
modal.find('#data-name').closest('tr')
```

to retrive the tr container.

djasmine:
Added wiki and readme.md. @yubowenok

yubowenok:
LGTM

djasmine:
There is still a front-end issue.
qq20160201-0 2x
The label at left side would contract when hiding some rows. How should I do with that. @yubowenok

yubowenok:
That is because the width of the first table column is detemined by the longest string in that colume. Previously "description" is the longest. I think you can try to set the first column to not wrap on whitespaces. Try css "white-space: nowrap".

djasmine:
Fixed the front-end issue. @yubowenok

yubowenok:
already LGTMed. You may merge.

templates/upload.html
```diff
@@ -9,13 +9,14 @@ <h4 class="modal-title">Upload Data</h4>
       <col width="100">
     </colgroup>
     <tr>
-      <td>Data Type</td>
+      <td style="white-space:nowrap">Data Type</td>
```
yubowenok, line 12: Please try to set this in css later. Inline css is to be avoided unless for initial property such as display none.

PR #87, opened 2016-02-02
fixed front-end issues in mapping upload

djasmine:
fixed frond-end issues in mapping upload

### DIFF
--- a/src/dialog.js
+++ b/src/dialog.js
@@ -292,6 +292,8 @@ genotet.dialog.upload_ = function() {
       var btnFile = modal.find('#btn-file');
       var fileDisplay = modal.find('#file-display');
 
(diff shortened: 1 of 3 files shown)


src/dialog.js
```diff
@@ -292,6 +292,8 @@ genotet.dialog.upload_ = function() {
       var btnFile = modal.find('#btn-file');
       var fileDisplay = modal.find('#file-display');
 
+      modal.find('tr:first').css('white-space', 'nowrap');
```
yubowenok: Can you add this to scss file? Static css settings go to files.

djasmine: Changed into dialog.scss.

yubowenok:
LGTM

PR #90, opened 2016-02-03
Fix network issues

djasmine:
fix node add/set/delete issues.
fix edge delete issues.
fix inconsistency of panel elements and the edge table.

### DIFF
--- a/src/components/network/network-loader.js
+++ b/src/components/network/network-loader.js
@@ -52,7 +52,9 @@ genotet.NetworkLoader.prototype.prepareGenes_ = function(inputGenes, isRegex) {
       }
     });
   } else {
-    genes = inputGenes.toLowerCase().split(/[\s,]+/);
+    genes = inputGenes.toLowerCase().split(/[\s,]+/).filter(function(gene) {
+        return gene in this.data.networkInfo.isTF;
+    }.bind(this));
   }
   return genes;
 };
@@ -168,7 +170,7 @@ genotet.NetworkLoader.prototype.addGenes_ = function(genes) {
   });
 
   if (!newGenes.length) {
-    genotet.warning('genes are already in the network');
+    genotet.warning('genes are already in the network or not existing');
     return;
   }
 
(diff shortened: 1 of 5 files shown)


src/components/network/network-loader.js
```diff
@@ -168,7 +170,7 @@ genotet.NetworkLoader.prototype.addGenes_ = function(genes) {
   });
 
   if (!newGenes.length) {
-    genotet.warning('genes are already in the network');
+    genotet.warning('genes are already in the network or not existing');
```
yubowenok: Can we check for each case and display "already in the network" and "not existing" separately?

src/components/network/network-table.js
```diff
@@ -209,3 +222,19 @@ genotet.NetworkTable.prototype.create = function(table, edges) {
   table.closest('#edge-list').css('width',
     /** @type {number} */(table.width()));
 };
+
+/**
+ * Change the table for the edge removal.
```
yubowenok: Updates the table for edge removal.

src/components/network/network-loader.js
djasmine: I am sorry the situation that genes are not existing is already warned by another function. So for here, just warning about already in the network.

src/components/network/network-table.js
djasmine: fixed

src/externs/datatables.js
```diff
@@ -26,7 +27,27 @@ jQuery.prototype.DataTable = function(opt_params) {};
...
-DataTables.prototype.invalidate = function(opt_params) {};
+DataTables.prototype.button = function(params) {};
+
+/**
+ * @param {Object=} opt_params
+ */
+DataTables.Api.prototype.invalidate = function(opt_params) {};
+
+/**
+ * @param {Object=} opt_params
```
yubowenok, line 40: I think this function has @return.

djasmine, line 40: Yes it should return DataTables.Api. But there will be issues when casting this data() into an array which is {!Array<!genotet.EdgeForTable>}. Maybe I could just add some annotation without casting coercively.

yubowenok, line 40: Can you typedef DataTables.Api to be Array so that it doesn't prevent your type coersion? Doc mentioned that DataTables.Api is Array-like https://datatables.net/reference/type/DataTables.Api

djasmine, line 40: Changed.

yubowenok:
LGTM

PR #92, opened 2016-02-06
Change reading TFA/Combination in text and fix issues

djasmine:
change reading TFA files in text
change reading combination in text
change uploaded fileName to the original fileName


src/css/dialog.css
```diff
@@ -0,0 +1,42 @@
+.modal-body td {
```
yubowenok: This file shouldn't be commited. This is a 3rd time this is happening. Could you please just check what you've commimted?

yubowenok: dialog.css.map isn't to be commited either.

server/expression.js
```diff
@@ -173,33 +173,34 @@ expression.readMatrix_ = function(buf) {
...
-  result.values = [];
-  for (var i = 0; i < n; i++) {
-    for (var j = 0; j < m; j++) {
-      var val = buf.readDoubleLE(offset);
-      offset += 8;
-      result.values.push(val);
-    }
+expression.readTFAmat_ = function(fileName) {
+  var lines = fs.readFileSync(fileName).toString().split('\n');
+  var parts = lines[0].split(/[\t\s]+/);
```
yubowenok: Please assume TFA to be with matrix text format. The original data was in R and I wrote a script to convert it to the .rawdata format you're parsing here. However the .rawdata format is now obsolete and we will switch to the tsv format you see for expression matrix. 

See my email for details.

server/uploader.js
```diff
@@ -30,29 +30,53 @@ uploader.ENTRY_SIZE_ = 12;
...
  * }} desc File description.
- * @param {Object} file File object received from multer.
+ * @param {!uploader.MulterFile} file File object received from multer.
  * @param {string} prefix The destination folder to upload the file to.
  * @param {string} bigWigToWigAddr Directory of script of UCSC bigWigToWig.
  * @return {Object} Success or not as a JS Object.
  */
 uploader.uploadFile = function(desc, file, prefix, bigWigToWigAddr) {
+  var fileName = file.originalname;
+  console.log('i am ' + file.originalname);
```
yubowenok: Please change "i am" to sth informative, or remove this log.

templates/upload.html
```diff
@@ -25,10 +25,6 @@ <h4 class="modal-title">Upload Data</h4>
       <td><input id="data-name" class="form-control"></td>
     </tr>
     <tr>
-        <td>File Name</td>
```
yubowenok, line 28: I think the front-end upload will be still sending fileName as an attr?

server/network.js
```diff
@@ -312,12 +311,7 @@ network.getIncidentEdges_ = function(file, gene) {
  */
 network.getComb_ = function(file, exp) {
```
yubowenok: Is this function working properly now? The other things have changed a lot. I think the front-end should be sending a list of node ids rather than an array of network.Nodes.

server/expression.js
```diff
@@ -211,13 +212,8 @@ expression.readTFAmat_ = function(buf) {
...
 expression.getTfaProfile_ = function(fileName, geneNames, conditionNames) {
-  var bufTfa = utils.readFileToBuf(fileName);
-  if (bufTfa == null) {
-    console.error('cannot read file', fileName);
-    return null;
-  }
-  var resultTfa = expression.readTFAmat_(/** @type {!Buffer} */(bufTfa));
-
+  var resultTfa = expression.readTFAmat_(fileName);
+  console.log(resultTfa.values.length);
```
yubowenok: logs need to be informative

server/uploader.js
```diff
@@ -30,29 +30,53 @@ uploader.ENTRY_SIZE_ = 12;
 uploader.DOUBLE_SIZE_ = 8;
 
 /**
+ * @typedef {{
+ *   filedname: string,
```
yubowenok: fileName

server/uploader.js
```diff
@@ -30,29 +30,53 @@ uploader.ENTRY_SIZE_ = 12;
...
+ *   originalname: string,
```
yubowenok: originalName

server/uploader.js
```diff
@@ -30,29 +30,53 @@ uploader.ENTRY_SIZE_ = 12;
...
+ *   encoding: string,
+ *   mimetype: string,
+ *   destination: string,
+ *   path: string,
+ *   filename: string,
```
yubowenok: already have fileName here.

server/uploader.js
yubowenok: filedname?

server/uploader.js
```diff
@@ -30,29 +30,53 @@ uploader.ENTRY_SIZE_ = 12;
...
+ *   size: number
+ * }}
+ */
+uploader.MulterFile;
```
yubowenok: Looks like this is to be written in server/externs/multer.js.

server/network.js
yubowenok: Let's rename this to getCombinedRegulation_ to match new code style.

yubowenok: Besides if the combined regulation is not working properly please fix in a separate PR (pullreq).

server/expression.js
yubowenok: See email. Sent you the new TSV with updated format like expression matrix.

djasmine:
You can review this pullreq again. @yubowenok

server/network.js
```diff
@@ -123,14 +123,12 @@ network.query.incidentEdges = function(query, networkPath) {
...
- * @return {!Array<!network.Node>}
+ * @return {!Array<string>}
  */
 network.query.combinedRegulation = function(query, networkPath) {
   var fileName = query.fileName;
-  var geneRegex = utils.decodeSpecialChar(query.geneRegex);
   var file = networkPath + fileName;
-  // TODO(jiaming): fix old file read
-  return network.getComb_(file, geneRegex);
+  return network.getCombination_(file, query.genes);
```
yubowenok: getCombinedRegulation_

src/components/network/network-loader.js
```diff
@@ -278,28 +278,26 @@ genotet.NetworkLoader.prototype.buildEdgeMap_ = function() {
...
-      addexp += '|'+ oexp;
-      loader.addNodes(addexp);
-    }
-  });
+/**
+ * Finds the combined regulation genes and adds them into the network.
+ * @param {string} inputGenes Input genes for combined regulation.
+ * @param {boolean} isRegex Whether combined regulation is based on regex.
+ */
+genotet.NetworkLoader.prototype.loadCombination = function(inputGenes,
```
yubowenok: loadCombinedRegulation.

src/components/network/network-panel.html
```diff
@@ -19,8 +19,14 @@
...
     <div class="item-title">Combined Regulations</div>
-    <button class="btn btn-default glyphicon glyphicon-refresh btn-update"></button>
-    <input class="form-control input-with-update">
+    <input class="form-control">
+    <div id="combined-input" class="input-checkbox">
+      <label class="radio-inline" name="string"><input type="radio" name="regulator-optradio" checked>text</label>
+      <label class="radio-inline" name="regex"><input type="radio" name="regulator-optradio">regex</label>
+    </div>
+    <div class="sub-buttons">
+      <button id="refresh-combined" class="btn btn-default btn-xs glyphicon glyphicon-refresh" title="refresh genes"></button>
```
yubowenok: id="refresh" is sufficient as the parent container already have "combined-regulation" as its id that indicates the purpose of this entire block.

src/components/network/network-panel.html
```diff
@@ -19,8 +19,14 @@
   </div>
   <div id="combined-regulation" class="panel-section">
     <div class="item-title">Combined Regulations</div>
-    <button class="btn btn-default glyphicon glyphicon-refresh btn-update"></button>
-    <input class="form-control input-with-update">
+    <input class="form-control">
+    <div id="combined-input" class="input-checkbox">
```
yubowenok: id="input" is sufficient as the parent container id already indicates "combined-regulation".

templates/upload.html
yubowenok, line 28: The front-end is still sending fileName in formData. But it seems that the backend is not expecting fileName to be within query body. So the front-end's formData.fileName could be redundant.

server/expression.js
yubowenok: Looks like this is still not reading the tsv format? @djasmine

yubowenok: If this function is no longer called you may remove it.

server/expression.js
```diff
@@ -211,38 +212,21 @@ expression.readTFAmat_ = function(buf) {
...
  */
 expression.getTfaProfile_ = function(fileName, geneNames, conditionNames) {
-  var bufTfa = utils.readFileToBuf(fileName);
-  if (bufTfa == null) {
-    console.error('cannot read file', fileName);
-    return null;
-  }
-  var resultTfa = expression.readTFAmat_(/** @type {!Buffer} */(bufTfa));
-
+  var result = expression.readExpression_(fileName, geneNames, conditionNames);
```
yubowenok: readMatrix could be a better name as we are sharing the read method.

server/expression.js
```diff
@@ -261,79 +245,6 @@ expression.getTfaProfile_ = function(fileName, geneNames, conditionNames) {
...
 
 /**
- * Gets the expression matrix data.
- * @param {string} file File name of the expression matrix data.
- * @param {string} exprows Regex selecting the genes.
- * @param {string} expcols Regex selecting the experiment conditions.
- * @return {?expression.Matrix}
- * @private
- */
-expression.getMatrix_ = function(file, exprows, expcols) {
```
yubowenok, line 271: Removal of this function seems to result in readMatrix_ being never used. You may remove readMatrix_ as well.

server/expression.js
yubowenok, line 137: Please add comments to indicate the reason we are calling readExpression here (Because the TFA profile has the same format as an expression matrix...etc).

yubowenok, line 137: Alternatively, you can write a readMatrix_ function which is shared by getTfaProfile_ and readExpression_ so that the code has better modularization.

yubowenok:
Done review @djasmine

djasmine:
You can review this pullreq again. @yubowenok

server/expression.js
```diff
@@ -19,7 +19,7 @@ function expression() {}
  *   numcols: number,
  *   rownames: !Array<string>,
  *   colnames: !Array<string>,
- *   values: !Array<!Array<number>>
+ *   values: !Array<number>
  * }}
  */
 expression.RawMatrix;
```
yubowenok: Looks like RawMatrix is never used and can be removed.

server/externs/multer.js
```diff
@@ -13,3 +13,17 @@ function multer(params) {}
...
+ *   originalname: string,
+ *   encoding: string,
+ *   mimetype: string,
+ *   destination: string,
+ *   path: string,
+ *   filename: string,
+ *   size: number
+ * }}
+ */
+multer.prototype.file;
```
yubowenok:
``` js
multer.File;
```

Uppercase the first letter indicates the type is a class-like or struct-like object. Lowercasing means the defined stuff is more like a namespace. If a class is also a namespace, you may lowercase it (e.g. multer).

If you have some struct (File) defined under a namespace (multer), it would be defined under the namespace instead of the namespace's prototype. For example, we have genotet.View rather than genotet.prototype.View. Prototype is like shared attributes across the instances of the same class (similar to "static" in C++ and Java). Therefore you would not have class defined under prototype. See prototype of JS for more details.

Please also add EOL.

server/uploader.js
```diff
@@ -34,25 +34,34 @@ uploader.DOUBLE_SIZE_ = 8;
  * @param {{
  *   type: string,
  *   name: string,
- *   fileName: string,
- *   description: string
+ *   description: string,
  * }} desc File description.
- * @param {Object} file File object received from multer.
+ * @param {!Object} file File object received from multer.
```
yubowenok: Would this be multer.File?

server/uploader.js
```diff
@@ -34,25 +34,34 @@ uploader.DOUBLE_SIZE_ = 8;
...
  * @param {string} bigWigToWigAddr Directory of script of UCSC bigWigToWig.
  * @return {Object} Success or not as a JS Object.
  */
 uploader.uploadFile = function(desc, file, prefix, bigWigToWigAddr) {
+  var fileName = file.originalname;
+  if (fs.existsSync(prefix + fileName) ||
+    fs.existsSync(prefix + fileName + '_chr')) {
+    return {
+      error: {
+        type: 'file already exists',
```
yubowenok: Usually type is a shorter keyword rather than a descriptive sentence, like 'fileExists'.

yubowenok:
LGTM with minor issues. You may merge after correcting them.

Issue #84, opened 2016-02-01
expression view improvements


yubowenok:
A few TODO tunings
- By default not show TFA
- Profile/TFA vertical axis ticks are a bit cluttered. Consider adjusting number of ticks based on the plot height.
- Zooming in could result in disappearing profiles. This is a bit confusing. I understand that you are restricting profiled genes to exist in the matrix. However I think since disappearing profiles are weird, it would be better to make them independent, so that profiled genes are kept intact when matrix row selection changes.
- If the user tries to add rows/columns that do not exist, it should show "no genes/conditions found" rather than "invalid input gene/condition selection".

PR #99, opened 2016-02-12
Implemented expression gene profiles improvements. Fixed #80 and parts of #84.

### LINKED ISSUE #80: expression panel text/regex radiobox shown incorrectly
snap

liana-yang:
#80 expression panel text/regex radiobox shown incorrectly
#84
- By default not show TFA
- Profile/TFA vertical axis ticks are a bit cluttered. Consider adjusting number of ticks based on the plot height.
- If the user tries to add rows/columns that do not exist, it should show "no genes/conditions found" rather than "invalid input gene/condition selection".

Added points to gene profiles and moved the interactions to points.

### DIFF
--- a/src/components/expression/expression-panel.js
+++ b/src/components/expression/expression-panel.js
@@ -113,8 +113,15 @@ genotet.ExpressionPanel.prototype.initPanel = function() {
     }.bind(this));
 
   // Input type update
+  var viewName = this.container.attr('id').replace('panel-view-', '');
+  this.container.find('#gene-input input')
+    .attr('name', viewName + '-gene-optradio');
+  this.container.find('#condition-input input')
+    .attr('name', viewName + '-condition-optradio');
+
   this.container.find('#gene-input label[name=regex] input')
     .click(function() {
+      console.log(this.container.attr('id').replace('panel-view-', ''));
       this.isGeneRegex_ = true;
     }.bind(this));
   this.container.find('#gene-input label[name=string] input')
@@ -143,7 +150,7 @@ genotet.ExpressionPanel.prototype.initPanel = function() {
 
       var geneNames = this.formatGeneInput(this.isGeneRegex_, geneInput);
       if (geneNames.length == 0) {
-        genotet.warning('invalid input gene selection');
+        genotet.warning('no genes found');
         return;
       }
       this.signal('update', {
@@ -168,7 +175,7 @@ genotet.ExpressionPanel.prototype.initPanel = function() {
       var conditionNames = this.formatConditionInput(this.isConditionRegex_,
         conditionInput);
       if (conditionNames.length == 0) {
-        genotet.warning('invalid input condition selection');
+        genotet.warning('no conditions found');
         return;
       }
       this.signal('update', {
@@ -217,9 +224,9 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
 /**
  * Adds the cell info into a given container.
  * @param {string} geneName Gene name of which info is to be displayed.
- * @param {string} conditionName Condition name of which info is to be
+ * @param {?|string} conditionName Condition name of which info is to be
  *     displayed.
- * @param {number} value Value of which info is to be displayed.
+ * @param {?|number} value Value of which info is to be displayed.
  * @param {!jQuery} container Info container.
  * @private
  */
@@ -233,6 +240,13 @@ genotet.ExpressionPanel.prototype.setCellInfo_ = function(geneName,
     .text(conditionName);
   container.children('#value').children('span')
     .text(value + '');
+
+  container.children('#condition').css('display', function() {
+    return conditionName ? 'inline' : 'none';
+  });
+  container.children('#number').css('display', function() {
+    return value != null ? 'inline' : 'none';
+  });
 };
 
 /**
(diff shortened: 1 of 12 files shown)


src/components/expression/expression-panel.js
```diff
@@ -113,8 +113,15 @@ genotet.ExpressionPanel.prototype.initPanel = function() {
...
   // Input type update
+  var viewName = this.container.attr('id').replace('panel-view-', '');
+  this.container.find('#gene-input input')
+    .attr('name', viewName + '-gene-optradio');
+  this.container.find('#condition-input input')
+    .attr('name', viewName + '-condition-optradio');
+
   this.container.find('#gene-input label[name=regex] input')
     .click(function() {
+      console.log(this.container.attr('id').replace('panel-view-', ''));
```
yubowenok: Please remove logs.

src/components/expression/expression-panel.js
```diff
@@ -113,8 +113,15 @@ genotet.ExpressionPanel.prototype.initPanel = function() {
     }.bind(this));
 
   // Input type update
+  var viewName = this.container.attr('id').replace('panel-view-', '');
```
yubowenok, line 117: If you have the view object reference you can get its name by view.name(). It's in general not a good idea to first encoding some thing into a string and then decode it back from the string. Preferrably we should get the value directly.

Possible TODO.

src/components/expression/expression-panel.js
```diff
@@ -217,9 +224,9 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
 /**
  * Adds the cell info into a given container.
  * @param {string} geneName Gene name of which info is to be displayed.
- * @param {string} conditionName Condition name of which info is to be
+ * @param {?|string} conditionName Condition name of which info is to be
```
yubowenok: undefined|string, or null|string. Question mark tells the compiler to NOT type-checking at all which is probably not what you want.

src/components/expression/expression-panel.js
```diff
@@ -233,6 +240,13 @@ genotet.ExpressionPanel.prototype.setCellInfo_ = function(geneName,
     .text(conditionName);
   container.children('#value').children('span')
     .text(value + '');
+
+  container.children('#condition').css('display', function() {
```
``` js
container.children('#condition').css('display', conditionName ? 'inline' : 'none');
```

You don't need a function here. Same below.

src/components/expression/expression-renderer.js
```diff
@@ -1037,58 +1043,99 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
...
-    .attr('d', function(profile) {
-      return line(heatmapData.values[profile.row]);
-    })
+  profileGroup.enter().append('g');
+  profileGroup
     .attr('transform', genotet.utils.getTransform([
       this.heatmapWidth_ / (heatmapData.conditionNames.length * 2),
       0
     ]))
+    .style('fill', function(profile) {
```
``` js
.style('fill', 'none');
```

src/components/expression/expression-renderer.js
```diff
@@ -1140,73 +1190,101 @@ genotet.ExpressionRenderer.prototype.drawTfaProfiles_ = function() {
...
-      var geneIndex = this.data.tfaGeneNameDict[tfaProfile.geneName];
-      return line(tfaData.tfaValues[geneIndex]);
-    }.bind(this))
+  profileGroup.enter().append('g');
+  profileGroup
     .attr('transform', genotet.utils.getTransform([
       this.heatmapWidth_ / (heatmapData.conditionNames.length * 2),
       this.profileHeight_
     ]))
+    .style('fill', function(tfaProfile) {
```
``` js
.style('fill', 'none');
```

src/components/expression/expression-renderer.js
```diff
@@ -1023,10 +1026,13 @@ genotet.ExpressionRenderer.prototype.drawGeneProfiles_ = function() {
     this.profileHeight_ - this.profileMargins_.bottom,
     yScaleTop
   ]).domain([heatmapData.valueMin, heatmapData.valueMax]);
+  var yAxisLength = this.profileHeight_ - this.profileMargins_.bottom -
```
yubowenok, line 1029: This is a comment for this entire function, not for this particular line.

This function drawGeneProfiles_ is getting way too chunky and it's even very hard to read the lines, as we have many nested d3 code. I suggest refactoring this function to multiple "smaller" functions (e.g. drawProfilePaths_, drawProfilePoints_).

TODO.

src/components/network/network-panel.js
```diff
@@ -45,6 +45,13 @@ genotet.NetworkPanel.prototype.initPanel = function() {
         }.bind(this));
   }, this);
 
+  // Input type update
+  var viewName = this.container.attr('id').replace('panel-view-', '');
```
yubowenok, line 50: ditto, suggest getting id directly.

Possible TODO.

yubowenok:
LGTM with minor issues.

src/components/expression/expression-panel.js
liana-yang, line 117: I agree with you. But how to access the view.name() from panel element?

src/components/expression/expression-renderer.js
liana-yang, line 1029: Added to TODO.

src/components/expression/expression-panel.js
yubowenok, line 117: You can store a reference to the view object when the view is creating panel. In other words, you can remember for each tab/panel the view object it corresponding to. This can be potentially useful later as well.

liana-yang, line 117: Okay. Added to TODO.

PR #115, opened 2016-02-20
enhance binding/bed rendering

### LINKED ISSUE #74: always show motif rects clearly
It's hard to see motif rects when labels are present. It's necessary that motifs rects are shown clearly regardless of the labels.

### LINKED ISSUE #96: prevent binging track height changes resulting from bed level-of-detail
The zoom level could result in 2 ways of rendering bed tracks (bed LoD). This could lead to immediate reduction of binding track height that makes it harder to eye-track binding data. Please prevent height change when the LoD of bed changes.

### LINKED ISSUE #106: binding view enhancements
- Display the data/file name for each track in the visualization.
- Highlight the track hovered in panel.
- Show coordinate difference (end - start) in the visualization.

### LINKED ISSUE #111: gap between system messages
image

liana-yang:
Fixed #74.
Fixed #96.
Fixed #106.
Fixed #111 .

### DIFF
--- a/src/components/base/view-renderer.js
+++ b/src/components/base/view-renderer.js
@@ -142,7 +142,8 @@ genotet.ViewRenderer.prototype.signal = function(eventType, opt_data) {
  * @param {number} canvasWidth Width of canvas.
  * @param {number} canvasHeight Height of canvas.
  */
-genotet.ViewRenderer.prototype.updateCanvasSize = function(canvasWidth,
+genotet.ViewRenderer.prototype.resizeCanvas = function(canvasWidth,
                                                            canvasHeight) {
+  this.canvas.attr('width', canvasWidth);
   this.canvas.attr('height', canvasHeight);
 };
--- a/src/components/base/view-renderer.js
+++ b/src/components/base/view-renderer.js
@@ -136,3 +136,13 @@ genotet.ViewRenderer.prototype.hideFailure = function() {
 genotet.ViewRenderer.prototype.signal = function(eventType, opt_data) {
   $(this).trigger('genotet.' + eventType, [opt_data]);
 };
+
+/**
+ * Sets the size of canvas.
+ * @param {number} canvasWidth Width of canvas.
+ * @param {number} canvasHeight Height of canvas.
+ */
+genotet.ViewRenderer.prototype.updateCanvasSize = function(canvasWidth,
+                                                           canvasHeight) {
+  this.canvas.attr('height', canvasHeight);
+};
--- a/src/components/base/view.js
+++ b/src/components/base/view.js
@@ -151,7 +151,7 @@ genotet.View.prototype.id = function() {
 /**
  * Sets the header text of the view. If null, return the current header.
  * @param {?string} headerText View header text.
- * @return {?string}
+ * @return {string}
  */
 genotet.View.prototype.headerText = function(headerText) {
   if (!headerText) {
@@ -160,13 +160,14 @@ genotet.View.prototype.headerText = function(headerText) {
   this.headerText_ = headerText;
   this.container.find('.header-text')
     .text(this.headerText_);
+  return this.headerText_;
 };
 
 /**
  * Sets header extra information of the view. If null, return the current
  * header.
  * @param {?string} extraInfo Extra information in the header.
- * @return {?string}
+ * @return {string}
  */
 genotet.View.prototype.headerExtraInfo = function(extraInfo) {
   if (!extraInfo) {
@@ -175,6 +176,7 @@ genotet.View.prototype.headerExtraInfo = function(extraInfo) {
   this.headerExtraInfo_ = extraInfo;
   this.container.find('.extra-info')
     .text(this.headerExtraInfo_);
+  return this.headerExtraInfo_;
 };
 
 /**
--- a/src/components/base/view.js
+++ b/src/components/base/view.js
@@ -255,7 +255,7 @@ genotet.View.prototype.signal = function(eventType, opt_data) {
  * @param {number} containerWidth Width of container.
  * @param {number} containerHeight Height of container.
  */
-genotet.View.prototype.updateContainerSize = function(containerWidth,
+genotet.View.prototype.resize = function(containerWidth,
                                                       containerHeight) {
   this.container.width(containerWidth);
   this.container.height(containerHeight);
--- a/src/components/base/view.js
+++ b/src/components/base/view.js
@@ -26,6 +26,9 @@ genotet.View = function(viewName) {
   /** @private {string} */
   this.headerText_ = '';
 
+  /** @private {string} */
+  this.headerExtraInfo_ = '';
+
   /** @private {boolean} */
   this.sendSignal_ = true;
 
@@ -160,6 +163,21 @@ genotet.View.prototype.headerText = function(headerText) {
 };
 
 /**
+ * Sets header extra information of the view. If null, return the current
+ * header.
+ * @param {?string} extraInfo Extra information in the header.
+ * @return {?string}
+ */
+genotet.View.prototype.headerExtraInfo = function(extraInfo) {
+  if (!extraInfo) {
+    return this.headerExtraInfo_;
+  }
+  this.headerExtraInfo_ = extraInfo;
+  this.container.find('.extra-info')
+    .text(this.headerExtraInfo_);
+};
+
+/**
  * Makes the view appear focused.
  * @param {boolean=} sendSignal Flag of sending signal.
  */
@@ -230,4 +248,15 @@ genotet.View.prototype.createPanel = function(container) {
  */
 genotet.View.prototype.signal = function(eventType, opt_data) {
   $(this).trigger('genotet.' + eventType, [opt_data]);
+};
+
+/**
+ * Sets the size of container.
+ * @param {number} containerWidth Width of container.
+ * @param {number} containerHeight Height of container.
+ */
+genotet.View.prototype.updateContainerSize = function(containerWidth,
+                                                      containerHeight) {
+  this.container.width(containerWidth);
+  this.container.height(containerHeight);
 };
--- a/src/components/binding/binding-loader.js
+++ b/src/components/binding/binding-loader.js
@@ -155,9 +155,12 @@ genotet.BindingLoader.prototype.loadBed = function(fileName, chr, xl, xr) {
     xl: xl,
     xr: xr
   };
+  var isAggregated = this.data.bed ? this.data.bed.aggregated : true;
   this.get(genotet.data.serverURL, params, function(data) {
     this.data.bed = data;
     this.data.bedName = fileName;
+    this.data.bed.aggregatedChanged = this.data.bed.aggregated != isAggregated;
+    console.log(this.data.bed.aggregatedChanged);
   }.bind(this), 'cannot load binding data');
 };
 
--- a/src/components/binding/binding-loader.js
+++ b/src/components/binding/binding-loader.js
@@ -155,7 +155,7 @@ genotet.BindingLoader.prototype.loadBed = function(fileName, chr, xl, xr) {
     xl: xl,
     xr: xr
   };
-  var isAggregated = this.data.bed ? this.data.bed.aggregated : true;
+  var isAggregated = !this.data.bed || this.data.bed.aggregated;
   this.get(genotet.data.serverURL, params, function(data) {
     this.data.bed = data;
     this.data.bedName = fileName;
--- a/src/components/binding/binding-loader.js
+++ b/src/components/binding/binding-loader.js
@@ -160,7 +160,6 @@ genotet.BindingLoader.prototype.loadBed = function(fileName, chr, xl, xr) {
     this.data.bed = data;
     this.data.bedName = fileName;
     this.data.bed.aggregatedChanged = this.data.bed.aggregated != isAggregated;
-    console.log(this.data.bed.aggregatedChanged);
   }.bind(this), 'cannot load binding data');
 };
 
--- a/src/components/binding/binding-panel.js
+++ b/src/components/binding/binding-panel.js
@@ -215,13 +215,20 @@ genotet.BindingPanel.prototype.updateFileListAfterLoading = function() {
     this.selectGenes_[index] = select;
     select.val(track.fileName).trigger('change');
 
-    // Set track fileName
-    select.on('select2:select', function(event) {
-      var fileName = event.params.data.id;
-      this.signal('updateTrack', {
-        trackIndex: index,
-        fileName: fileName
-      });
-    }.bind(this));
+    // Set track fileName and highlight the track hovered in panel.
+    select
+      .on('select2:select', function(event) {
+        var fileName = event.params.data.id;
+        this.signal('updateTrack', {
+          trackIndex: index,
+          fileName: fileName
+        });
+      }.bind(this))
+      .off().on('select2:open', function(event) {
+        this.signal('highlightTrack', index);
+      }.bind(this))
+      .on('select2:close', function(event) {
+        this.signal('unhighlightTrack', index);
+      }.bind(this));
   }, this);
 };
--- a/src/components/binding/binding-panel.js
+++ b/src/components/binding/binding-panel.js
@@ -216,6 +216,7 @@ genotet.BindingPanel.prototype.updateFileListAfterLoading = function() {
     select.val(track.fileName).trigger('change');
 
     // Set track fileName and highlight the track hovered in panel.
+    var listOpened = false;
     select
       .on('select2:select', function(event) {
         var fileName = event.params.data.id;
@@ -225,10 +226,22 @@ genotet.BindingPanel.prototype.updateFileListAfterLoading = function() {
         });
       }.bind(this))
       .on('select2:open', function(event) {
+        listOpened = true;
         this.signal('highlightTrack', index);
       }.bind(this))
       .on('select2:close', function(event) {
+        listOpened = false;
         this.signal('unhighlightTrack', index);
+      }.bind(this));
+    ui.find('.selection span')
+      .on('mouseenter', function(event) {
+        console.log('in');
+        this.signal('highlightTrack', index);
+      }.bind(this))
+      .on('mouseleave', function(event) {
+        if (!listOpened) {
+          this.signal('unhighlightTrack', index);
+        }
       }.bind(this));
   }, this);
 };
--- a/src/components/binding/binding-panel.js
+++ b/src/components/binding/binding-panel.js
@@ -224,7 +224,7 @@ genotet.BindingPanel.prototype.updateFileListAfterLoading = function() {
           fileName: fileName
         });
       }.bind(this))
-      .off().on('select2:open', function(event) {
+      .on('select2:open', function(event) {
         this.signal('highlightTrack', index);
       }.bind(this))
       .on('select2:close', function(event) {
--- a/src/components/binding/binding-renderer.js
+++ b/src/components/binding/binding-renderer.js
@@ -34,6 +34,12 @@ genotet.BindingRenderer = function(container, data) {
    */
   this.bedRectHeight_ = 0;
 
+  /**
+   * Height of a single bed unit. This is re-computed upon resize event.
+   * @private {number}
+   */
+  this.bedUnitHeight_ = 0;
+
   // Navigation state.
   /** @private {!Array<number>} */
   this.zoomTranslate_ = [0, 0];
@@ -120,8 +126,7 @@ genotet.BindingRenderer.prototype.TRACK_NAME_TEXT_HEIGHT_ = 10;
 /**
  * When there are more than this limit number of exons, we draw exons as
  * abstract rectangles.
- * @private
- * @type {number}
+ * @private {number}
  */
 genotet.BindingRenderer.prototype.EXON_ABSTRACT_LIMIT_ = 200;
 
@@ -427,6 +432,16 @@ genotet.BindingRenderer.prototype.layout = function() {
     .attr('transform', function(track, index) {
       return genotet.utils.getTransform([0, this.detailHeight_ * index]);
     }.bind(this));
+
+  // Set up bed unit height and rect height.
+  var bedData = this.bedPositions_;
+  this.bedRectHeight_ = (this.bedHeight_ - this.BED_MARGINS_.TOP -
+    this.BED_MARGINS_.BOTTOM) / bedData.length;
+  this.bedUnitHeight_ = this.bedRectHeight_;
+  if (!this.data.bed.aggregated && this.data.options.showBedLabels) {
+    this.bedRectHeight_ -= this.BED_LABEL_SIZE_;
+  }
+  this.bedRectHeight_ = Math.max(this.bedRectHeight_, this.BED_MIN_HEIGHT_);
 
   // Set up bed shallow line.
   var lineParameters = {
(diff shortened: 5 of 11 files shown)


src/components/binding/binding-loader.js
```diff
@@ -155,9 +155,11 @@ genotet.BindingLoader.prototype.loadBed = function(fileName, chr, xl, xr) {
     xl: xl,
     xr: xr
   };
+  var isAggregated = this.data.bed ? this.data.bed.aggregated : true;
```
yubowenok: Consider 

``` js
var isAggregated = !this.data.bed || this.data.bed.aggregated;
```

src/components/binding/binding-panel.js
```diff
@@ -215,13 +215,20 @@ genotet.BindingPanel.prototype.updateFileListAfterLoading = function() {
...
+    // Set track fileName and highlight the track hovered in panel.
+    select
+      .on('select2:select', function(event) {
+        var fileName = event.params.data.id;
+        this.signal('updateTrack', {
+          trackIndex: index,
+          fileName: fileName
+        });
+      }.bind(this))
+      .off().on('select2:open', function(event) {
```
yubowenok: Will this be offing the select event above?

I think highlighting is supposed to be triggered on hovering rather than list opening?

src/components/binding/binding-renderer.js
```diff
@@ -110,6 +110,12 @@ genotet.BindingRenderer.prototype.OVERVIEW_HEIGHT = 25;
 genotet.BindingRenderer.prototype.BED_HEIGHT = 25;
 /** @const {number} */
 genotet.BindingRenderer.prototype.BED_SHALLOW_WIDTH = 1;
+/** @const {number} */
```
yubowenok: All private or protected.

src/components/binding/binding.js
```diff
@@ -275,3 +283,16 @@ genotet.BindingView.prototype.updateTrackWithMapping_ = function(fileName) {
...
+
+/**
+ * Updates coordinate difference in header.
+ * @param {number} coordinateDifference Coordinate difference of binding
+ * tracks.
+ * @private
+ * */
+genotet.BindingView.prototype.updateCoordinateDifferenceInHearder_ =
+  function(coordinateDifference) {
+    this.container.find('.extra-info')
```
yubowenok: header is under the control of view.js and any change to header content is preferred to go through view API.

yubowenok:
Please add a meaningful name for this PR. Thanks.

src/components/binding/binding-renderer.js
```diff
@@ -725,7 +773,7 @@ genotet.BindingRenderer.prototype.drawBed_ = function() {
       return rectWidth < this.BED_MIN_WIDTH ? this.BED_MIN_WIDTH :
         rectWidth;
     }.bind(this))
-    .attr('height', this.bedRectHeight_ > 0 ? this.bedRectHeight_ : 0)
+    .attr('height', Math.max(this.bedRectHeight_, this.BED_MIN_HEIGHT))
```
yubowenok: The rendering code should avoid computing size values. this.bedRectHeight_ can actually be the final height applied and can be computed preferrably in layout().

src/components/binding/binding-renderer.js
```diff
@@ -774,12 +822,29 @@ genotet.BindingRenderer.prototype.updateDetailHeight_ = function() {
   var exonsHeight = this.data.options.showExons ?
     this.EXON_HEIGHT : 0;
   if (this.data.bed.aggregated) {
+    if (this.data.bed.aggregatedChanged) {
+      this.canvasHeight -= this.bedHeight_;
```
yubowenok, line 842: This is a rendering function and layouting could be done in dedicated layout().

src/components/binding/binding-renderer.js
```diff
@@ -774,12 +822,29 @@ genotet.BindingRenderer.prototype.updateDetailHeight_ = function() {
...
+      var containerHeight = this.container.height() - this.bedHeight_;
+      this.container.height(containerHeight);
+      this.canvas.attr('height', this.canvasHeight);
```
yubowenok: Canvas size should be changed via View API rather than renderer. Renderer is supposed to have read-only access to view size.

src/components/binding/binding-renderer.js
```diff
@@ -774,12 +822,29 @@ genotet.BindingRenderer.prototype.updateDetailHeight_ = function() {
...
+    }
     this.bedHeight_ = this.BED_HEIGHT;
   } else {
-    this.bedHeight_ = this.data.options.showBed ?
-    (this.canvasHeight - exonsHeight - overviewHeight) *
-    this.BED_HEIGHT_PROPORTION : 0;
+    if (this.data.bed.aggregatedChanged) {
+      this.bedHeight_ = this.data.options.showBed ?
+      this.bedHeight_ + this.BED_HEIGHT_EXTENSION : 0;
```
yubowenok: Incorrect indentation.

src/components/binding/binding-panel.js
liana-yang: We should set the highlighting under this open event because we only want to trigger the highlighting when we open the binding file list select2. There are many select2s in the page and they are set as a kind of same block in body tag. If we do not do this, we will listen all the select2 hover events by jquery.

liana-yang: off() is necessary because the listener will react multiple times due to unknown reasons from select2. We discussed this long long ago.

yubowenok: "because we only want to trigger the highlighting when we open the binding file list select2" -> false. I think the requirement is highlight on hovering select list. The highlighting is for the user to observe which track corresponds to which select. Clicking is one more unnecessary perceptional gap.

yubowenok: No we didn't discuss anything about using off(). I only remember that you somehow fixed it by upgrading the select2 version. Will off() just remove the added handler of "select2:select"?

src/components/binding/binding-renderer.js
```diff
@@ -96,45 +96,54 @@ genotet.BindingRenderer = function(container, data) {
...
+genotet.BindingRenderer.prototype.TRACK_NAME_HEIGHT_ = 50;
+/** @private @const {number} */
+genotet.BindingRenderer.prototype.TRACK_NAME_TEXT_OFFSET_ = 10;
+/** @private @const {number} */
+genotet.BindingRenderer.prototype.TRACK_NAME_TEXT_HEIGHT_ = 10;
 
 /**
  * When there are more than this limit number of exons, we draw exons as
  * abstract rectangles.
+ * @private
```
yubowenok: @private {number} is sufficient

src/components/expression/expression-renderer.js
```diff
@@ -217,54 +217,41 @@ genotet.ExpressionRenderer.ZoomStatus = function(params) {
...
+genotet.ExpressionRenderer.prototype.DEFAULT_PROFILE_LEGEND_HEIGHT_ = 25;
+/** @private @const {number} */
+genotet.ExpressionRenderer.prototype.DEFAULT_PROFILE_LEGEND_MARGIN_ = 10;
+/** @private @const {number} */
+genotet.ExpressionRenderer.prototype.DEFAULT_PROFILE_MARGIN_ = 40;
+/** @private @const {number} */
+genotet.ExpressionRenderer.prototype.DEFAULT_PROFILE_CIRCLE_SIZE_ = 1;
+
+/** @private @const {number} */
+genotet.ExpressionRenderer.prototype.DEFAULT_HEATMAP_GRADIENT_HEIGHT_ = 20;
```
yubowenok: It's unlikely that gradient height gets changed. DEFAULT could be saved.

src/components/expression/expression-renderer.js
```diff
@@ -217,54 +217,41 @@ genotet.ExpressionRenderer.ZoomStatus = function(params) {
...
-
-/** @const {number} */
+/** @private @const {number} */
+genotet.ExpressionRenderer.prototype.DEFAULT_PROFILE_LEGEND_HEIGHT_ = 25;
+/** @private @const {number} */
+genotet.ExpressionRenderer.prototype.DEFAULT_PROFILE_LEGEND_MARGIN_ = 10;
+/** @private @const {number} */
+genotet.ExpressionRenderer.prototype.DEFAULT_PROFILE_MARGIN_ = 40;
+/** @private @const {number} */
+genotet.ExpressionRenderer.prototype.DEFAULT_PROFILE_CIRCLE_SIZE_ = 1;
```
yubowenok: It's unlikely that CIRCLE_SIZE can be changed. DEFAULT could be saved.

src/components/expression/expression-renderer.js
```diff
@@ -217,54 +217,41 @@ genotet.ExpressionRenderer.ZoomStatus = function(params) {
...
-
-/** @const {number} */
-genotet.ExpressionRenderer.prototype.LABEL_DIFFERENCE = 10;
-
-/** @const {number} */
-genotet.ExpressionRenderer.prototype.HEATMAP_LEGEND_MARGIN = 1;
-
-/** @const {number} */
+/** @private @const {number} */
+genotet.ExpressionRenderer.prototype.DEFAULT_PROFILE_LEGEND_HEIGHT_ = 25;
```
yubowenok: If you are not planning for custom legend font size (which I personally think is not necessary), DEFAULT can also be saved here. You may want to also revise other DEFAULT's.

src/components/binding/binding-renderer.js
liana-yang, line 842: This is a function used to update detail height......

src/components/binding/binding-panel.js
liana-yang: I mean, all the select2 open list (no matter in binding panel or expression panel or network panel) will be integrated into a same block like: 

``` html
<span class="select2-container select2-container--default select2-container--open" style="position: absolute; top: 121px; left: 719.203px;">
    <span class="select2-dropdown select2-dropdown--below" dir="ltr" style="width: 252px;">
        <span class="select2-search select2-search--dropdown">
            <input class="select2-search__field" type="search" tabindex="0" autocomplete="off" autocorrect="off" autocapitalize="off" spellcheck="false" role="textbox">
        </span>
        <span class="select2-results">
            <ul class="select2-results__options" role="tree" id="select2-1562-results" aria-expanded="true" aria-hidden="false">
                <li class="select2-results__option select2-results__option--highlighted" role="treeitem" aria-selected="true">expressionMatrix (expressionMatrix)</li>
                <li class="select2-results__option" role="treeitem" aria-selected="false">rnaseq (rnaseq)</li>
            </ul>
        </span>
    </span>
</span>
```

They have all same attributes. If we did not restrict the hover listener to the binding list selection list in binding panel, we would trigger this event when any select2 in the page is open and hovered, so that we can not decide which one is from binding file selection list.

yubowenok: That is not true. You can always select a specific element. The select2 container created is NOT global but instead local in panel so you should be able to select a single select2 container down from the panel, and then add event listeners to that specific select container. "we would trigger this event when any select2 in the page" is because of incorrect implementation.

yubowenok:
Please add a meaningful name for this PR. @Shauro

src/components/binding/binding-renderer.js
yubowenok, line 842: Sorry. I misread the function title due to code collapsing.

src/components/binding/binding-panel.js
liana-yang: No, they are global. You can check it in your browser.

yubowenok: image

What is global is the dropdown list. However we are not supposed to do anything with hovering the dropdown list. The highlighting is on when hovering the select container inside the panel.

yubowenok: These are supposed to trigger highlighting:
image

yubowenok: Instead of this:
image

src/components/base/view.js
```diff
@@ -160,6 +163,21 @@ genotet.View.prototype.headerText = function(headerText) {
 };
 
 /**
+ * Sets header extra information of the view. If null, return the current
+ * header.
+ * @param {?string} extraInfo Extra information in the header.
+ * @return {?string}
```
yubowenok: Maybe we can return always return a string. If it is set info operation, then return extraInfo itself. Same for headerText().

src/components/base/view.js
```diff
@@ -231,3 +249,14 @@ genotet.View.prototype.createPanel = function(container) {
 genotet.View.prototype.signal = function(eventType, opt_data) {
   $(this).trigger('genotet.' + eventType, [opt_data]);
 };
+
+/**
+ * Sets the size of container.
+ * @param {number} containerWidth Width of container.
+ * @param {number} containerHeight Height of container.
+ */
+genotet.View.prototype.updateContainerSize = function(containerWidth,
```
yubowenok: You can call it resize(), if not conflicting with other methods.

src/components/base/view-renderer.js
```diff
@@ -136,3 +136,13 @@ genotet.ViewRenderer.prototype.hideFailure = function() {
 genotet.ViewRenderer.prototype.signal = function(eventType, opt_data) {
   $(this).trigger('genotet.' + eventType, [opt_data]);
 };
+
+/**
+ * Sets the size of canvas.
+ * @param {number} canvasWidth Width of canvas.
+ * @param {number} canvasHeight Height of canvas.
+ */
+genotet.ViewRenderer.prototype.updateCanvasSize = function(canvasWidth,
```
yubowenok: function naming: resizeCanvas

Preferrably resizeCanvas could go to view API as well. Any specific reason for putting it under renderer?

src/components/base/view-renderer.js
```diff
@@ -136,3 +136,13 @@ genotet.ViewRenderer.prototype.hideFailure = function() {
...
+                                                           canvasHeight) {
+  this.canvas.attr('height', canvasHeight);
```
yubowenok, line 148: If the method has canvasWidth as input then it should set attr('width') as well.

src/components/binding/binding-panel.js
yubowenok: Besides I still think that when you call off() the following handler will be cancelled:

``` js
.on('select2:select', function(event) {
...
}
```

Is that so? Is select2:select still being fired?

src/components/binding/binding-renderer.js
```diff
@@ -96,45 +102,53 @@ genotet.BindingRenderer = function(container, data) {
...
+genotet.BindingRenderer.prototype.BED_MIN_WIDTH_ = 3;
+/** @private @const {number} */
+genotet.BindingRenderer.prototype.BED_MIN_HEIGHT_ = 3;
+/** @private @const {number} */
+genotet.BindingRenderer.prototype.BED_HEIGHT_EXTENSION_ = 200;
+
+/** @private @const {!Array<number>} */
+genotet.BindingRenderer.ZOOM_EXTENT_ = [1, 20000000];
+/** @private @const {number} */
+genotet.BindingRenderer.ZOOM_INTERVAL_ = 500;
```
yubowenok: define this on prototype

src/components/binding/binding-renderer.js
```diff
@@ -96,45 +102,53 @@ genotet.BindingRenderer = function(container, data) {
...
+genotet.BindingRenderer.prototype.BED_LABEL_SIZE_ = 13;
+/** @private @const {number} */
+genotet.BindingRenderer.prototype.BED_MIN_WIDTH_ = 3;
+/** @private @const {number} */
+genotet.BindingRenderer.prototype.BED_MIN_HEIGHT_ = 3;
+/** @private @const {number} */
+genotet.BindingRenderer.prototype.BED_HEIGHT_EXTENSION_ = 200;
+
+/** @private @const {!Array<number>} */
+genotet.BindingRenderer.ZOOM_EXTENT_ = [1, 20000000];
```
yubowenok: define this on prototype too, using function return.

src/components/binding/binding-renderer.js
```diff
@@ -770,22 +829,52 @@ genotet.BindingRenderer.prototype.drawBed_ = function() {
...
   if (this.data.bed.aggregated) {
-    this.bedHeight_ = this.BED_HEIGHT;
+    if (this.data.bed.aggregatedChanged) {
+      this.canvasHeight -= this.bedHeight_;
+      var containerHeight = this.container.height() - this.bedHeight_;
+      this.signal('updateContainerSize', {
+        containerWidth: this.container.width(),
+        containerHeight: containerHeight
+      });
+      this.signal('updateCanvasSize', {
```
yubowenok: 'resizeCanvas' is preferred

src/components/binding/binding-renderer.js
```diff
@@ -770,22 +829,52 @@ genotet.BindingRenderer.prototype.drawBed_ = function() {
...
+      this.bedHeight_ = this.data.options.showBed ?
+        this.bedHeight_ + this.BED_HEIGHT_EXTENSION_ : 0;
+      this.canvasHeight += this.BED_HEIGHT_EXTENSION_;
+      var containerHeight = this.container.height() +
+        this.BED_HEIGHT_EXTENSION_;
+      this.signal('updateContainerSize', {
+        containerWidth: this.container.width(),
+        containerHeight: containerHeight
+      });
+      this.signal('updateCanvasSize', {
```
yubowenok: When the resize event is fired, will the view be re-rendered/re-layouted AFTER the resize? What is the re-rendering/re-layouting mechanism involved?

src/components/binding/binding-panel.js
liana-yang: I see. It is my mistake firing the event on the list. Thx!

liana-yang: It has been cancelled... off() should not be used here.

src/components/base/view.js
liana-yang: They aren't same now?

src/components/base/view-renderer.js
liana-yang: Because this.canvas is defined in renderer.

yubowenok: Ok. Please excuse my memory :)

src/components/binding/binding-renderer.js
liana-yang: why using function return?

src/components/binding/binding-renderer.js
liana-yang: Yes, they will. Zoom in/out -> updateDetailHeight -> resize the view if necessary -> re-layouting -> re-rendering.

src/components/binding/binding-renderer.js
yubowenok: Otherwise you would get compiler error saying It cannot have non-primitive
type as zoom extent is an array.

liana-yang: I see.

src/components/base/view.js
yubowenok:
``` js
/**
 * ...
 * @return {string}
 */
genotet.View.prototype.headerExtraInfo = function(extraInfo) {
   if (!extraInfo) {
     return this.headerExtraInfo_;
   }
   this.headerExtraInfo_ = extraInfo;
   this.container.find('.extra-info')
    .text(this.headerExtraInfo_);
   return this.headerExtraInfo_;
 };
```

src/components/network/network-renderer.js
```diff
@@ -90,26 +91,28 @@ genotet.NetworkRenderer.MouseState = {
...
-/** @const {number} */
-genotet.NetworkRenderer.prototype.NODE_LABEL_OFFSET_X = 10;
-/** @const {number} */
-genotet.NetworkRenderer.prototype.NODE_LABEL_OFFSET_Y =
-    genotet.NetworkRenderer.prototype.NODE_LABEL_SIZE / 2;
-/** @const {number} */
-genotet.NetworkRenderer.prototype.NODE_SIZE = 6;
-/** @const {number} */
-genotet.NetworkRenderer.prototype.EDGE_ARROW_LENGTH = 10;
+genotet.NetworkRenderer.ZOOM_EXTENT_ = [.03125, 8];
```
yubowenok: please do prototype and function return.

src/components/binding/binding-panel.js
```diff
@@ -215,13 +215,20 @@ genotet.BindingPanel.prototype.updateFileListAfterLoading = function() {
...
+    // Set track fileName and highlight the track hovered in panel.
+    select
+      .on('select2:select', function(event) {
+        var fileName = event.params.data.id;
+        this.signal('updateTrack', {
+          trackIndex: index,
+          fileName: fileName
+        });
+      }.bind(this))
+      .on('select2:open', function(event) {
```
yubowenok, line 228: I think this will require one mouse click. However I think "hovering" means mouseover, without mouse clicking.

liana-yang, line 228: Does open event need trigger highlighting? I think it is necessary as well because the user may want to know which track they are going to change while they are selecting lists.

yubowenok, line 228: Yes you're right. So I guess it's both hovering and clicking.

src/components/base/view.js
```diff
@@ -231,3 +251,14 @@ genotet.View.prototype.createPanel = function(container) {
...
   $(this).trigger('genotet.' + eventType, [opt_data]);
 };
+
+/**
+ * Sets the size of container.
+ * @param {number} containerWidth Width of container.
+ * @param {number} containerHeight Height of container.
+ */
+genotet.View.prototype.resize = function(containerWidth,
+                                                      containerHeight) {
```
yubowenok: The indentation is a little weird here. Suggest aligning containerHeight with containerWidth.

src/components/binding/binding-panel.js
```diff
@@ -215,13 +215,33 @@ genotet.BindingPanel.prototype.updateFileListAfterLoading = function() {
...
+        listOpened = true;
+        this.signal('highlightTrack', index);
+      }.bind(this))
+      .on('select2:close', function(event) {
+        listOpened = false;
+        this.signal('unhighlightTrack', index);
+      }.bind(this));
+    ui.find('.selection span')
+      .on('mouseenter', function(event) {
+        console.log('in');
```
yubowenok: Please remove log.

yubowenok:
Please sync.

src/components/binding/binding-renderer.js
```diff
@@ -144,7 +163,7 @@ genotet.BindingRenderer.prototype.init = function() {
 
   // Set up the detailed histogram zoom.
   this.zoom_ = d3.behavior.zoom()
-    .scaleExtent(genotet.BindingRenderer.ZOOM_EXTENT)
+    .scaleExtent(this.ZOOM_EXTENT_)
```
yubowenok: this.ZOOM_EXTENT_()

src/components/base/view.js
yubowenok: @Shauro

src/components/network/network-renderer.js
```diff
@@ -118,7 +123,7 @@ genotet.NetworkRenderer.prototype.init = function() {
 
   /** @private {d3.zoom} */
   this.zoom_ = d3.behavior.zoom()
-    .scaleExtent(genotet.NetworkRenderer.ZOOM_EXTENT)
+    .scaleExtent(this.ZOOM_EXTENT_)
```
yubowenok: this.ZOOM_EXTENT_()

PR #140, opened 2016-02-26
Add start script and fix file not found log

djasmine:
Add start.sh script.
Add fileName to file not found error log.
Fix minor bug of reading binding file.


yubowenok:
Sync? LGTM

Issue #141, opened 2016-02-26
create server.log for timestamped server logging


yubowenok:

djasmine:
fixed

PR #143, opened 2016-02-26
Add timestamp logs and data format doc

djasmine:
Add timestamp to server logs #141 .
Add docs for data format #142.


server/bed.js
```diff
@@ -3,6 +3,7 @@
  */
 
 var fs = require('fs');
+var utils = require('./utils');
```
yubowenok: logger would be better encapsulated by a class of its own.

``` js
var log = require('./log');
log.write(...);
```

server/externs/express.js
```diff
@@ -53,6 +53,15 @@ express.Response.prototype.status = function(data) {};
 express.Response.prototype.send = function(msg, code) {};
 
 /**
+ * @constructor
```
yubowenok, line 50: Where is this used?

server/utils.js
```diff
@@ -78,3 +79,17 @@ utils.decodeSpecialChar = function(url) {
...
+
+/**
+ * Prints logs with timestamps.
+ * @param {!Array<string>} args Content to be logged.
+ */
+utils.serverLog = function(args) {
+  if (!args.length) {
+    return;
+  }
+  var date = new Date();
```
yubowenok: It will be better to write both the posix time (int) and the server local time (date string).

src/menu.js
```diff
@@ -43,9 +43,6 @@ genotet.menu.init = function() {
   $('#about').click(function() {
     genotet.dialog.create('about');
   });
-  $('#doc').click(function() {
```
yubowenok: How to open the doc if this is removed?

server/bed.js
yubowenok: To diff from client log, you may say

``` js
log.serverLog(content);
log.clientLog(user, session, content);
```

djasmine:

server/bed.js
```diff
@@ -66,8 +68,10 @@ bed.query.motifs = function(query, bedPath) {
   var chr = query.chr;
   var dir = bedPath + fileName + '_chr/' + fileName + '_chr' + chr;
   if (!fs.existsSync(dir)) {
+    var error = 'bed file ' + fileName + ' not found.';
+    log.serverLog([error]);
```
yubowenok: Umm using an array of strings could be counter-intuitive. Consider a function with variable-length argument, such as console.log, genotet.error (front-end).

src/preset.js
```diff
@@ -31,7 +31,7 @@ genotet.preset.NETWORK_PARAMS_ = {
  */
 genotet.preset.EXPRESSION_PARAMS_ = {
   fileName: 'rnaseq',
-  tfaFileName: 'tfa.mat.tsv',
+  tfaFileName: 'tfa_rnaseq.mat.tsv',
```
yubowenok: what's the naming rule? I think rnqseq doesn't have tfa, which I remember to be associated with b.subtilis network.

server/binding.js
```diff
@@ -102,8 +103,10 @@ binding.query.histogram = function(query, bindingPath) {
     '.bcwig';
   var descriptionPath = bindingPath + fileName + '.desc';
   if (!fs.existsSync(file)) {
+    var error = 'binding file ' + fileName + ' not found.';
+    log.serverLog(error);
     return {
```
yubowenok, line 108: I think this should be

``` js
return error;
```

Same for similar places throughout the server code.

yubowenok:
tests broken

server/binding.js
djasmine, line 108: Here error is just a string, and we would look at data.error after it.

yubowenok, line 108: Oh, okay.

djasmine:

doc.html
```diff
@@ -8,225 +8,75 @@
   <link type="text/css" rel="stylesheet" href="dist/doc.css" />
```
yubowenok: This page needs some proper styling. Right now it is a bit too old fashioned.

doc.html
```diff
@@ -8,225 +8,75 @@
...
+    <h2> Genotet - Help Document </h2>
+    <p>For more information, contact <a href="mailto:bowen.yu@nyu.edu">Bowen</a>, <a href="mailto:jiamingdong@nyu.edu">Jiaming</a> or <a href="mailto:ling.yang@nyu.edu">Liana</a>. </p>
+    <p>Note: This document is currently only for data format of data uploading.</p>
+    <h3>Network</h3>
+    <ul>
+      <li>Currently, we accept *.tsv file for network uploading.</li>
+      <li>The first line contains N + 2 space seperated strings, containing 2 table head of "regulator", "target" and N names of values.</li>
+      <li>Following M lines, containing M edges from regulators to targets, each line contains N + 2 elements. They are 1 regulator, 1 target and N float values.</li>
+      <li>It is like
+      <pre>regulator target value_name1 ... value_nameN</pre>
```
yubowenok: If it's pre you can put all lines in one pre block.

``` html
<pre>
  GENE condition[1] ... condition[N]
  gene[1] value[1][1] ... value[1][N]
  ...
  gene[M] value[M][1] ... value[M][N]
</pre>
```

doc.html
yubowenok: You may include the bootstrap css easily to make the style compatible with the main system.
Take a look at the templates: http://getbootstrap.com/getting-started/#examples-framework

djasmine:

doc.html
```diff
@@ -4,230 +4,90 @@
...
+    <h2> Genotet - Help Document </h2>
+    <p>For more information, contact <a href="mailto:bowen.yu@nyu.edu">Bowen</a>, <a href="mailto:jiamingdong@nyu.edu">Jiaming</a> or <a href="mailto:ling.yang@nyu.edu">Liana</a>. </p>
+    <p>Note: This document is currently only for data format of data uploading.</p>
+    <h3>Network</h3>
+    <ul>
+      <li>Currently, we accept *.tsv file for network uploading.</li>
+      <li>The first line contains N + 2 space seperated strings, containing 2 table head of "regulator", "target" and N names of values.</li>
+      <li>Following M lines, containing M edges from regulators to targets, each line contains N + 2 elements. They are 1 regulator, 1 target and N float values.</li>
+      <li>It is like
+      <pre>
```
yubowenok: Could you please correct the indentation here? pre need to be indented 2 more spaces than its parent li. The closing </pre> need to be on a new line.

yubowenok:
LGTM with minor indentation issue.

PR #150, opened 2016-03-11
Add user log

djasmine:
Add user log function.
Call user log in other function.
Write down user log in server.
Fix bugs.

### DIFF
--- a/server/log.js
+++ b/server/log.js
@@ -12,6 +12,26 @@ module.exports = log;
  */
 function log() {}
 
+/** @const */
+log.query = {};
+
+/**
+ * @typedef {{
+ *   timestamp: number,
+ *   type: string,
+ *   action: string
+ * }}
+ */
+log.UserLog;
+
+/**
+ * @typedef {{
+ *   userName: string,
+ *   logs: !Array<!log.UserLog>
+ * }}
+ */
+log.query.UserLogList;
+
 /**
  * Prints logs with timestamps.
  * @param {...*} var_args Content to be logged.
@@ -23,23 +43,18 @@ log.serverLog = function(var_args) {
   var date = new Date();
   var content = '[' + dateFormat(date, 'yyyy-mm-dd_HH:MM:ss') + '_' +
     date.getTime() + '] ';
-  for (var i = 0; i < arguments.length; i++) {
-    content += arguments[i];
-  }
+  content += arguments.slice(1).join('_');
   console.log(content);
 };
 
 /**
  * Prints user activity logs.
  * @param {string} userPath Path to the user folder.
- * @param {string} userName User name to be logged.
- * @param {!Array<{
- *   timestamp: number,
- *   type: string,
- *   action: string
- * }>} logs Log information.
+ * @param {!log.query.UserLogList} query Query parameters.
  */
-log.userLog = function(userPath, userName, logs) {
+log.userLog = function(userPath, query) {
+  var logs = query.logs;
+  var userName = query.userName;
   if (!logs.length) {
     return;
   }
--- a/server/log.js
+++ b/server/log.js
@@ -28,3 +28,31 @@ log.serverLog = function(var_args) {
   }
   console.log(content);
 };
+
+/**
+ * Prints user activity logs.
+ * @param {string} userPath Path to the user folder.
+ * @param {string} userName User name to be logged.
+ * @param {!Array<{
+ *   timestamp: number,
+ *   type: string,
+ *   action: string
+ * }>} logs Log information.
+ */
+log.userLog = function(userPath, userName, logs) {
+  if (!logs.length) {
+    return;
+  }
+  var folder = userPath + userName + '/';
+  var date = new Date();
+  var logFile = folder + 'log_' + dateFormat(date, 'yyyy-mm-dd_HH') + '.log';
+  var fd = fs.openSync(logFile, 'w+');
+  logs.forEach(function(log) {
+    var logString = '';
+    var date = new Date(log.timestamp);
+    logString += '[' + dateFormat(date, 'yyyy-mm-dd_HH:MM:ss') + ']\t';
+    logString += log.type + ' ' + log.action;
+    fs.writeSync(fd, logString + '\n');
+  });
+  fs.closeSync(fd);
+};
--- a/server/log.js
+++ b/server/log.js
@@ -19,8 +19,7 @@ log.query = {};
 /**
  * @typedef {{
  *   timestamp: number,
- *   type: string,
- *   action: string
+ *   content: string
  * }}
  */
 log.UserLog;
@@ -69,8 +68,8 @@ log.userLog = function(userPath, query) {
     var logString = '';
     date.setMilliseconds(log.timestamp);
     logString += '[' + dateFormat(date, 'yyyy-mm-dd_HH:MM:ss') + '_' +
-      log.timestamp + ']\t';
-    logString += log.type + ' ' + log.action;
+      log.timestamp + '] ';
+    logString += log.content;
     fs.writeSync(fd, logString + '\n');
   });
   fs.closeSync(fd);
--- a/server/log.js
+++ b/server/log.js
@@ -63,7 +63,7 @@ log.userLog = function(userPath, query) {
   }
   var date = new Date();
   var logFile = folder + 'log_' + dateFormat(date, 'yyyy-mm-dd_HH') + '.log';
-  var fd = fs.openSync(logFile, 'w+');
+  var fd = fs.openSync(logFile, 'a');
   logs.forEach(function(log) {
     var logString = '';
     date.setMilliseconds(log.timestamp);
--- a/server/log.js
+++ b/server/log.js
@@ -27,7 +27,7 @@ log.UserLog;
 
 /**
  * @typedef {{
- *   userName: string,
+ *   username: string,
  *   logs: !Array<!log.UserLog>
  * }}
  */
@@ -57,11 +57,8 @@ log.serverLog = function(var_args) {
  */
 log.userLog = function(userPath, query) {
   var logs = query.logs;
-  var userName = query.userName;
-  if (!logs.length) {
-    return;
-  }
-  var folder = userPath + userName + '/log/';
+  var username = query.username;
+  var folder = userPath + username + '/log/';
   if (!fs.existsSync(folder)) {
     fs.mkdirSync(folder);
   }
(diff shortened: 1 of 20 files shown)


djasmine:
Note: for now, I am confused with the zooming functions for all the views, I would add them after discussing with Liana later.

server/log.js
```diff
@@ -12,6 +13,26 @@ module.exports = log;
  */
 function log() {}
 
+/** @const */
+log.query = {};
+
+/**
+ * @typedef {{
+ *   timestamp: number,
```
liana-yang, line 21: timestamp seems not only contain number. It should be string or date?

src/components/binding/binding-loader.js
```diff
@@ -36,6 +36,7 @@ genotet.BindingLoader.prototype.LOCUS_MARGIN_RATIO = .1;
  */
 genotet.BindingLoader.prototype.load = function(fileName, bedName, chr,
                                                 opt_track) {
+  genotet.logger.log('binding', 'load', fileName, bedName, chr, opt_track);
```
liana-yang: I think log trackIndex here is better than opt_track.

server/log.js
```diff
@@ -12,6 +13,26 @@ module.exports = log;
...
+ *   type: string,
+ *   action: string
```
liana-yang: action seems optional? And sometimes it contains not only one parameter?

liana-yang:
LGTM

src/components/expression/expression-panel.js
```diff
@@ -193,6 +200,7 @@ genotet.ExpressionPanel.prototype.initPanel = function() {
   this.container.find('#out').click(function() {
     if (this.data.zoomStack.length > 0) {
       var zoomStatus = this.data.zoomStack.pop();
+      genotet.logger.log('expression', 'expressionZoomOut', zoomStatus);
```
yubowenok: expressionZoomOut seems to be redundant? will zoomOut be sufficient?

What is the result of logging a zoomStatus object?

src/log.js
```diff
@@ -0,0 +1,68 @@
...
+  genotet.logger.logList.push({
+    timestamp: date.getTime(),
+    type: type,
+    action: action
+  });
+  console.log(genotet.logger.logList.length);
+  if (genotet.logger.logList.length == genotet.logger.MAX_SIZE_) {
+    var params = {
+      logs: genotet.logger.logList,
+      userName: 'anonymous', // will call api of user system
```
yubowenok: The username shall be retrieved at the server side with authentication.

src/log.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Adds log to the log list, returns if collecting enough messages.
+ * @param {...*} var_args
+ */
+genotet.logger.log = function(var_args) {
+  if (!arguments.length || arguments.length < 2) {
+    return;
+  }
+  var type = arguments[0];
```
yubowenok: Looks like log type would better be an enum.

src/log.js
```diff
@@ -0,0 +1,68 @@
...
+      logs: genotet.logger.logList,
+      userName: 'anonymous', // will call api of user system
+      type: 'user-log'
+    };
+    $.ajax({
+      url: genotet.data.userURL,
+      type: 'POST',
+      data: params,
+      dataType: 'json'
+    }).done(function(data) {
```
yubowenok: Why empty functions?

src/log.js
```diff
@@ -0,0 +1,68 @@
...
+    action: action
+  });
+  console.log(genotet.logger.logList.length);
+  if (genotet.logger.logList.length == genotet.logger.MAX_SIZE_) {
+    var params = {
+      logs: genotet.logger.logList,
+      userName: 'anonymous', // will call api of user system
+      type: 'user-log'
+    };
+    $.ajax({
```
yubowenok: $.post for short.

src/data.js
```diff
@@ -31,6 +31,12 @@ genotet.data.serverURL;
 genotet.data.uploadURL;
 
 /**
+ * User queries for user info and log.
+ * @type {string}
+ */
+genotet.data.userURL;
```
yubowenok: This looks like a weird name. Why not logURL and authURL?

src/log.js
```diff
@@ -0,0 +1,68 @@
...
+      url: genotet.data.userURL,
+      type: 'POST',
+      data: params,
+      dataType: 'json'
```
yubowenok: What kind of json data are you retrieving here?

server/log.js
```diff
@@ -24,7 +45,35 @@ log.serverLog = function(var_args) {
...
+/**
+ * Prints user activity logs.
+ * @param {string} userPath Path to the user folder.
+ * @param {!log.query.UserLogList} query Query parameters.
+ */
+log.userLog = function(userPath, query) {
+  var logs = query.logs;
+  var userName = query.userName;
+  if (!logs.length) {
+    return;
```
yubowenok: Under what circumstances could logs.length be zero?

server/log.js
```diff
@@ -24,7 +45,35 @@ log.serverLog = function(var_args) {
...
 };
+
+/**
+ * Prints user activity logs.
+ * @param {string} userPath Path to the user folder.
+ * @param {!log.query.UserLogList} query Query parameters.
+ */
+log.userLog = function(userPath, query) {
+  var logs = query.logs;
+  var userName = query.userName;
```
yubowenok: username is a whole word.

server/log.js
```diff
@@ -24,7 +45,35 @@ log.serverLog = function(var_args) {
...
+  if (!fs.existsSync(folder)) {
+    fs.mkdirSync(folder);
+  }
+  var date = new Date();
+  var logFile = folder + 'log_' + dateFormat(date, 'yyyy-mm-dd_HH') + '.log';
+  var fd = fs.openSync(logFile, 'w+');
+  logs.forEach(function(log) {
+    var logString = '';
+    date.setMilliseconds(log.timestamp);
+    logString += '[' + dateFormat(date, 'yyyy-mm-dd_HH:MM:ss') + ']\t';
```
yubowenok: timestamp?

server/log.js
yubowenok, line 21: timestamp is strictly a number.

server/server.js
```diff
@@ -136,11 +145,28 @@ var upload = multer({
  */
 var exonFile = bindingPath + 'exons.bin';
 
+app.use(bodyParser.urlencoded({extended: true}));
+app.use(bodyParser.json());
+
 /**
- * POST request is not used as it conflicts with jsonp.
+ * User info POST handler.
```
yubowenok: This description doesn't match the handler content. It actually handles user-log, but I didn't see user info?

src/log.js
```diff
@@ -0,0 +1,68 @@
...
+    action += ' ' + arguments[i];
+  }
+  var date = new Date();
+  genotet.logger.logList.push({
+    timestamp: date.getTime(),
+    type: type,
+    action: action
+  });
+  console.log(genotet.logger.logList.length);
+  if (genotet.logger.logList.length == genotet.logger.MAX_SIZE_) {
```
yubowenok: Logs should also be sent when the browser page is closed.

src/log.js
djasmine: Oh because I am confused about what should we do here. Because there must be some error in the back-end, should we show genotet.error here?

src/log.js
djasmine: The POST params is an Object, should be send as json.

src/log.js
djasmine: How can I catch the close action?

src/components/binding/binding-loader.js
djasmine: Where can I get trackIndex?

src/log.js
yubowenok: Hmm I think the messages are irrelevant to the user. But if the transmission failed the system log should try to send the log again some time later.

src/log.js
yubowenok: Oh I see. Will this be the same as:

``` js
$.post(genotet.data.userURL, params, 'json');
```

src/log.js
yubowenok: window.onbeforeunload?

src/data.js
djasmine: I think I would combine user log and user authentication, so this would be the same as that branch?

yubowenok: I don't think it is a good idea to combine logging and authentication as they are two separate things. We can have one URL for authentication and another for logging.

djasmine:
You can review again. @Shauro @yubowenok

server/server.js
```diff
@@ -136,11 +145,28 @@ var upload = multer({
...
 /**
- * POST request is not used as it conflicts with jsonp.
+ * User log POST handler.
+ */
+app.post('/genotet/log', function(req, res) {
+  log.serverLog('POST', req.body.type);
+
+  var prefix = userPath;
+  switch (req.body.type) {
+    case 'user-log':
```
yubowenok: If the request is posted to dedicated url like '/genotet/log', then it might not be necessary to further include req.body.type, because anything posted to this url must be user-log.

src/genotet.js
```diff
@@ -59,6 +63,7 @@ genotet.init = function() {
 genotet.warning = function(var_msgs) {
   var msg = Array.prototype.slice.call(arguments).join(' ');
   console.warn(msg);
+  genotet.logger.log('warning', 'msg');
```
yubowenok: I think 'msg' should be msg rather than the string 'msg'?

yubowenok: same for the occurrences below.

src/genotet.js
```diff
@@ -9,6 +9,10 @@ $(document).ready(function() {
   genotet.init();
 });
 
+$(window).on('beforeunload', function() {
```
yubowenok: It's not a good idea to write global level code. You may add an init function in log such as genotet.logger.init and call that init() in the genotet system initialization (genotet.init).

djasmine:

src/components/binding/binding-loader.js
```diff
@@ -36,6 +36,7 @@ genotet.BindingLoader.prototype.LOCUS_MARGIN_RATIO = .1;
  */
 genotet.BindingLoader.prototype.load = function(fileName, bedName, chr,
                                                 opt_track) {
+  genotet.logger.log('binding', 'load', fileName, bedName, chr);
```
yubowenok: Use enum for all the log types.

``` js
/** @enum {string} */
genotet.logger.Type = {
  BINDING: 'binding',
  ...
};

/**
 * @param {genotet.logger.Type} type
 * ...
 */
genotet.logger.log = function(type, ...) {
};

// Usage
genotet.logger.log(genotet.logger.Type.BINDING, ...);
```

yubowenok:
LGTM

PR #151, opened 2016-03-13
Fixed add/remove gene profile significant lag.

### LINKED ISSUE #146: add/remove gene profile significant lag on remote server

liana-yang:
Fixed #146: add/remove gene profile significant lag on remote server.


src/externs/select2.js
```diff
@@ -27,6 +27,15 @@ select2.Event.prototype.params;
  */
 jQuery.prototype.select2 = function(opt_params) {};
 
+/** @type {?} */
```
yubowenok: question mark needs to be avoided as much as possible. Arguments would be necessary here to explain why you want the compiler to not check this code.

src/components/expression/expression-panel.js
```diff
@@ -213,16 +205,57 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
...
-  this.updateGeneSelections();
+
+  if (genes.length == 0) {
+    return;
+  }
+
+  $.fn.select2.amd.require(['select2/data/array', 'select2/utils'],
+    function(ArrayData, Utils) {
+      /**
+       * @suppress {missingProperties}
```
yubowenok: This is way too hacky.

src/components/expression/expression-panel.js
```diff
@@ -213,16 +205,57 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
...
+       */
+      var CustomData = function($element, options) {
+        CustomData.__super__.constructor.call(this, $element, options);
+      };
+      /** @type {{Extend:function()}} */(Utils.Extend(CustomData, ArrayData));
```
yubowenok: Preferrably type annotation with record type is expanded into multiple lines.

src/components/expression/expression-panel.js
```diff
@@ -213,16 +205,57 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
...
-  this.selectProfiles_ = profile.select2({
-    data: genes,
-    multiple: true
-  });
-  this.container.find('#profile .select2-container').css({
-    width: '100%'
-  });
-  this.updateGeneSelections();
+
+  if (genes.length == 0) {
```
yubowenok: length

src/components/expression/expression-panel.js
```diff
@@ -224,16 +218,66 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
...
+        /**
+         * @type {{
+         *   __super__: function(?, ?)
+         * }}
+         */(CustomData).__super__.constructor.call(
+          this, $element, options);
+      };
+      /**
+       * @type {{
+       *   Extend:function()
```
yubowenok: Missing a space after Extend:

src/components/expression/expression-panel.js
```diff
@@ -224,16 +218,66 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
...
+       * }}
+       */(Utils.Extend(CustomData, ArrayData));
+
+      var pageSize = genotet.ExpressionPanel.prototype.PROFILE_PAGE_SIZE_;
+      CustomData.prototype.query = function(params, callback) {
+        if (!('page' in params)) {
+          params.page = 1;
+        }
+        var data = {};
```
``` js
callback({
  results: ...,
  pagination: {
    more: ...
  }
});
```

src/components/expression/expression-panel.js
```diff
@@ -224,16 +218,66 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
...
+          this, $element, options);
+      };
+      /**
+       * @type {{
+       *   Extend:function()
+       * }}
+       */(Utils.Extend(CustomData, ArrayData));
+
+      var pageSize = genotet.ExpressionPanel.prototype.PROFILE_PAGE_SIZE_;
+      CustomData.prototype.query = function(params, callback) {
```
yubowenok: can you annotate params?

src/components/expression/expression-panel.js
```diff
@@ -224,16 +218,66 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
...
-  });
-  this.updateGeneSelections();
+
+  if (!genes.length) {
+    return;
+  }
+
+  $.fn.select2.amd.require(['select2/data/array', 'select2/utils'],
+    function(ArrayData, Utils) {
+      var CustomData = function($element, options) {
```
yubowenok: why do we need a dollar sign in the variable name? Does it have special indication? if not you may just use "element".

src/components/expression/expression-panel.js
```diff
@@ -224,16 +218,66 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
...
+         */(CustomData).__super__.constructor.call(
+          this, $element, options);
+      };
+      /**
+       * @type {{
+       *   Extend:function()
+       * }}
+       */(Utils.Extend(CustomData, ArrayData));
+
+      var pageSize = genotet.ExpressionPanel.prototype.PROFILE_PAGE_SIZE_;
```
yubowenok: var pageSize = this.PROFILE_PAGE_SIZE_;

src/components/expression/expression-panel.js
```diff
@@ -224,16 +218,66 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
...
+        var data = {};
+        data.results = genes.slice((params.page - 1) * pageSize,
+          params.page * pageSize);
+        data.pagination = {};
+        data.pagination.more = params.page * pageSize < genes.length;
+        callback(data);
+      };
+
+      this.selectProfiles_ = profile.select2({
+        ajax: {},
```
yubowenok: What's the purpose of empty ajax?

src/components/expression/expression-panel.js
```diff
@@ -224,16 +218,66 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
...
-  this.container.find('#profile .select2-container').css({
-    width: '100%'
-  });
-  this.updateGeneSelections();
+
+  if (!genes.length) {
+    return;
+  }
+
+  $.fn.select2.amd.require(['select2/data/array', 'select2/utils'],
```
yubowenok: I think you don't need to require this every time you updateGenes? CustomData only need to be setup once and you can use it later on by

``` js
this.selectProfiles_ = profile.select2({
  dataAdapter: CustomData
  ...
});
```

src/components/expression/expression-panel.js
```diff
@@ -224,16 +218,66 @@ genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
...
+        callback(data);
+      };
+
+      this.selectProfiles_ = profile.select2({
+        ajax: {},
+        dataAdapter: CustomData,
+        multiple: true
+      });
+
+      if (!this.profileSelectionIsListened_) {
```
yubowenok: Try **not** to use a flag just for checking whether event listeners have been set. There should be other ways in which you can guarantee you set up the listeners only once.

src/externs/select2.js
```diff
@@ -27,6 +27,15 @@ select2.Event.prototype.params;
  */
 jQuery.prototype.select2 = function(opt_params) {};
 
+/** @type {Object} */
```
yubowenok: Though Object is typed, amd is declared to be a function which does not cope with the type annotation. You probably want a namespace.

src/components/expression/expression-panel.js
liana-yang: $element is according to the [select2 document](https://select2.github.io/announcements-4.0.html). I think it is consistent with the source code of select2 and we should keep it.

src/components/expression/expression-panel.js
liana-yang: It is used to query data for select2 and it should be kept empty in this situation according to the document.

yubowenok: ack

src/components/expression/expression-panel.js
yubowenok: ack

src/components/expression/expression-panel.js
liana-yang: According to the document, the require is asynchronously and CustomData should be used under the require. And we should use CustomData every time we updateGenes, so that we have to require this every time correspondingly.

yubowenok: You can require it once, wait for the async require to be done, and then
use CustomData later on. This is like registering a service and a same
service won't need to be registered twice. I don't see why a same require
need to be performed multiple times.

If you are **100% sure** that this require needs to be done every time you
use it, then it's actually weird antipattern and you can send me the
documentation. I will investigate more closely.

liana-yang: I tried your approach and it doesn't work. The require is used to only require the parts of Select2 that you need. I think it is necessary every time when you set up the select2. When you update select2, you reset it and should require again. [select2 document](https://select2.github.io/announcements-4.0.html) is here.

yubowenok: "I tried your approach and it doens't work"... Hmm, I don't want to write code :/, but anyway 

Add the below to the bottom of genotet.utils

``` js
genotet.utils.SELECT2_PAGE_SIZE_ = 20;

/**
 * @param {!Element} $element
 * @param {!Object} options
 * @constructor
 */
genotet.utils.CustomData = function($element, options) {
  /**
   * @type {{
   *   __super__: function(?, ?)
   * }}
   */(genotet.utils.CustomData).__super__.constructor.call(
    this, $element, options);
};

$.fn.select2.amd.require(['select2/data/array', 'select2/utils'],
  function(ArrayData, Utils) {
  var CustomData =
  /**
   * @type {{
   *   Extend: function(?, ?)
   * }}
   */(Utils).Extend(genotet.utils.CustomData, ArrayData);

  /**
   * The query of CustomData.
   * @param {!Object<{
   *   page: number
   * }>} params Parameters for the query of CustomData.
   * @param {function(?)} callback Callback function.
   */
  CustomData.prototype.query = function(params, callback) {
    var pageSize = genotet.utils.SELECT2_PAGE_SIZE_;
    if (!('page' in params)) {
      params.page = 1;
    }
    var genes = this.options.options.data;
    callback({
      results: genes.slice((params.page - 1) * pageSize,
        params.page * pageSize),
      pagination: {
        more: params.page * pageSize < genes.length
      }
    });
  };
});
```

Then edit the updateGenes so that it looks like

``` js
/**
 * Updates the genes in the profile list.
 * Select2 will regenerate the selection list each time updated.
 * @param {!Array<string>} gene List of genes.
 */
genotet.ExpressionPanel.prototype.updateGenes = function(gene) {
  var genes = gene.map(function(gene) {
    return {
      id: gene,
      text: gene
    };
  });
  var profile = this.container.find('#profile select').empty();

  if (!genes.length) {
    return;
  }

  this.selectProfiles_ = profile.select2({
    data: genes,
    ajax: {},
    dataAdapter: genotet.utils.CustomData,
    multiple: true
  });

  this.container.find('#profile .select2-container').css({
    width: '100%'
  });
  this.updateGeneSelections();
};
```

And it should work the same as your version. CustomData is a service and require sets up the service so that it can be reused later. You just need to have a way to reference to the service, and pass your data into the service.

yubowenok: A few more comments:
- You may want to rename "genotet.utils.CustomData" to be something more meaningful, such as genotet.utils.Select2PageAdapter.
- Currently the way that select2 works with paging is very hard to use. You can't even search for a gene with known name. More particularly, if the user types in some letters, we need to show only those items that match this search. As far as I know, you can capture what the user types by "params.term" in CustomData.prototype.query. You can then filter the genes by the keywords.
- Another problem is that right now because of the data adapter I don't even see the search box.
  image
  It was something like this before
  image
  so that I can see what I type and search for something.

liana-yang: I tried your codes above. In this function, an error said 'genotet.utils.CustomData.**super** is undefined'. Are you sure CustomData can be defined out of amd function?

``` js
/**
 * @param {!Element} $element
 * @param {!Object} options
 * @constructor
 */
genotet.utils.CustomData = function($element, options) {
  /**
   * @type {{
   *   __super__: function(?, ?)
   * }}
   */(genotet.utils.CustomData).__super__.constructor.call(
    this, $element, options);
};
```

yubowenok: I believe this

``` js
(Utils).Extend(genotet.utils.CustomData, ArrayData);
```

will give it all the necessary properties.

yubowenok: I actually run this code long ago and verified that it worked. If you see that it doesn't now, then first make sure you followed the steps correctly. If the problem persists, point me to your branch on this.

liana-yang: I checked the select2 [document](https://select2.github.io/examples.html) and found that Select2 multi-value select boxes doesn't have search box. I haven't found illustrations about setting search box in Select2 multi-value select boxes yet.

yubowenok: IIRC when you type-in any character you can only list those items matching the characters. That's equivalent to a search box.

Has the above issue been resolved?

liana-yang: Yes, above issues are solved except for the search matching issue.

yubowenok: I think efficient searching is non-trivial. For the simplest try brute-force search for the moment. Make another PR for the search.

yubowenok:
LGTM

src/components/expression/expression-panel.js
liana-yang: Okay. I will start another PR soon.

Issue #155, opened 2016-03-20
use one data path


yubowenok:
Right now we have this in config

```
networkPath = .../genotet_data/network/
bindingPath = .../genotet_data/binding/
expressionPath = .../genotet_data/expression/
bigWigToWigPath = .../genotet_data/bigWigToWig
uploadPath = .../genotet_data/upload/
bedPath = .../genotet_data/bed/
mappingPath = .../genotet_data/mapping/
```

But it seems to me that this is getting over complicated. I suggest we just do

```
dataPath = .../genotet_data/
```

and then we control all the stuffs under this folder, put them in respective folders.

djasmine:
Done.

Issue #154, opened 2016-03-18
gene input not working properly in network


yubowenok:
- create a network with empty input gene
- type in any text/regex gene input and click "+" (add gene)

```
genotet.js:186 Uncaught TypeError: Cannot read property 'nodes' of null
```

djasmine:
Done.

Issue #153, opened 2016-03-18
real upload progress bar


yubowenok:
please also bug fix
- upload error message not visible, blocked by modal dialog
- modal dialog shall auto close when upload succeeds

djasmine:
upload part done, need to put processing progress into db.

djasmine:
Done.

PR #156, opened 2016-03-22
Change to real upload progress

djasmine:
Fix issues of no input on creating network. #154
Change to real upload progress bar. #153 
Change to one data Path and one script Path. #155


server/network.js
```diff
@@ -470,9 +470,11 @@ network.listNetwork_ = function(networkPath) {
  */
 network.incrementalEdges_ = function(file, genes, nodes) {
   var oldNodes = {};
-  nodes.forEach(function(node) {
-    oldNodes[node.id] = true;
-  });
+  if (nodes.length) {
```
yubowenok: if nodes.length == 0, then nodes.forEach will not enter the inner function. the if condition is thus redundant?

server/server.js
```diff
@@ -100,36 +100,26 @@ function config() {
...
-        break;
     }
   }
+  bindingPath = dataPath + 'wiggle/';
+  networkPath = dataPath + 'network/';
+  expressionPath = dataPath + 'expression/';
+  uploadPath = dataPath + 'upload/';
+  bedPath = dataPath + 'bed/';
+  mappingPath = dataPath + 'mapping/';
+  userPath = dataPath + 'user/';
```
yubowenok: I think we are not using userPath? @djasmine @Shauro

server/uploader.js
```diff
@@ -199,7 +199,7 @@ uploader.bigWigToBCWig = function(prefix, bwFile, bigWigToWigAddr, uploadPath) {
     var fd = fs.openSync(uploadPath + bwFile + '.finish', 'w');
     fs.writeSync(fd, 'finish');
     fs.closeSync(fd);
-    log.serverLog(['binding data separate done.']);
+    log.serverLog('binding data separate done.');
```
yubowenok: separation done / separated

server/uploader.js
```diff
@@ -281,7 +281,7 @@ uploader.bedSort = function(prefix, bedFile, uploadPath) {
 uploader.checkFinish = function(query, prefix) {
   var isFinish = fs.existsSync(prefix + query.fileName + '.finish');
   if (isFinish) {
-    fs.unlink(prefix + query.fileName + '.finish');
+    fs.unlinkSync(prefix + query.fileName + '.finish');
```
yubowenok, line 298: Please add a TODO to mark this to be added to DB.

src/dialog.js
```diff
@@ -426,6 +426,10 @@ genotet.dialog.upload_ = function() {
         formData.append('description',
           /** @type {string} */(modal.find('#description').val()));
         formData.append('file', file[0].files[0]);
+        var uploadPercent = 100;
+        if (fileType == 'bed' || fileType == 'binding') {
```
yubowenok: genotet.FileType.BED, genotet.FileType.BINDING

src/dialog.js
```diff
@@ -435,58 +439,81 @@ genotet.dialog.upload_ = function() {
...
             } else {
               genotet.success('data uploaded');
             }
           })
           .fail(function(res) {
+            modal.modal('hide');
             genotet.error('failed to upload data');
           });
-        genotet.dialog.uploadProgress_(fileName);
+        if (fileType == 'bed' || fileType == 'binding') {
```
yubowenok: ditto

src/dialog.js
```diff
@@ -435,58 +439,81 @@ genotet.dialog.upload_ = function() {
...
+    };
+    $.get(genotet.data.serverURL, params, function(data) {
+      if (data) {
+        clearInterval(interval);
+        modal.modal('hide');
+      }
     });
+    modal.find('.progress').children('.progress-bar')
+      .css('width', widthPercent);
+  }, genotet.dialog.queryInterval_);
```
yubowenok: Is query interval a constant? If so it should be 

``` js
genotet.dialog.QUERY_INTERVAL_ = ...;
```

src/dialog.js
```diff
@@ -435,58 +439,81 @@ genotet.dialog.upload_ = function() {
...
+    number++;
+    // This is a fake progress bar.
+    // Increase 1% for the progress bar every 3 seconds.
+    // When reaching 99, do not increase it any more.
+    var widthPercent = Math.min(number, 99) + '%';
+    var params = {
+      type: 'check-finish',
+      fileName: fileName
+    };
+    $.get(genotet.data.serverURL, params, function(data) {
```
yubowenok: function(isFinished) might be clearer

src/dialog.js
yubowenok: Will 3*1000 be too long for the interval? It takes 3secs \* 30percent = 90 secs. As far as I remember processing doesn't take that long?

test/config
```diff
@@ -1,7 +1,2 @@
-bindingPath = dist/data/wiggle/
-networkPath = dist/data/network/
-expressionPath = dist/data/expression/
-bedPath = dist/data/bed/
-uploadPath = dist/data/upload/
-bigWigToWigPath = test/bigWigToWig
-mappingPath = dist/data/mapping/
+dataPath = dist/data/
+bigWigToWigPath = test/bigWigToWig
```
yubowenok: EOL

yubowenok:
LGTM with minor issues

server/network.js
djasmine: It will break when nodes.length == 0, when the original network is empty.

server/server.js
djasmine: Because I am using this path to store the logs, it might be a logPath?

yubowenok: Ok  then log path could be better

yubowenok: If we have a folder for user logs, then it could be 'log/username/logFiles'

server/network.js
yubowenok: Really? What is the reason that it breaks? From the code it seems to work exactly the same way.

djasmine: The server will return an error if the nodes array is empty.

yubowenok: What error is it? I don't get it... If nodes = [], the the forEach has no effect, so that oldNodes is an empty object as it is, and everything will be the same?

djasmine:

server/bed.js
```diff
@@ -47,23 +47,28 @@ bed.MotifsResult;
...
-// Start public APIs
-/**
- * @param {!bed.query.Motifs} query
+ * }} query
  * @param {string} bedPath
  * @return {bed.MotifsResult|bed.Error}
  */
 bed.query.motifs = function(query, bedPath) {
+  if (query.fileName == undefined) {
+    return {
```
yubowenok: Since the response only has one attribute, it is preferred to inline it for shorter code.

``` js
{error: '...'}
```

server/binding.js
```diff
@@ -143,6 +143,11 @@ binding.query.exons = function(query, exonFile) {
  * }|binding.Error}
  */
 binding.query.locus = function(query, exonFile) {
+  if (query.gene == undefined) {
+    return /** @type {binding.Error} */({
```
yubowenok: Why does this need type coersion, while the others do not?

server/bed.js
```diff
@@ -47,23 +47,28 @@ bed.MotifsResult;
...
-
-// Start public APIs
-/**
- * @param {!bed.query.Motifs} query
+ * }} query
  * @param {string} bedPath
  * @return {bed.MotifsResult|bed.Error}
  */
 bed.query.motifs = function(query, bedPath) {
+  if (query.fileName == undefined) {
```
yubowenok: Use === (triple-equal) to compare with undefined, so as to differ from comparing with null. Sometimes null has different meanings.

server/log.js
```diff
@@ -52,12 +51,26 @@ log.serverLog = function(var_args) {
...
+  }
+  if (query.logs == undefined) {
+    return {
+      error: 'logs is undefined'
+    };
+  }
   var logs = query.logs;
   var username = query.username;
-  var folder = userPath + username + '/log/';
+  var folder = userPath + username + '/logFiles/';
```
yubowenok: I think /logs/ should be okay? logFiles sounds a bit verbose.

server/log.js
```diff
@@ -52,12 +51,26 @@ log.serverLog = function(var_args) {
...
  * @param {string} userPath Path to the user folder.
- * @param {!log.query.UserLogList} query Query parameters.
+ * @param {*|{
+ *   username: string,
+ *   logs: !Array<!log.UserLog>
+ * }} query Query parameters.
+ * @return {log.Error}
  */
-log.userLog = function(userPath, query) {
+log.query.userLog = function(userPath, query) {
```
yubowenok: should this be logPath?

server/log.js
yubowenok: If it is logPath, then logPath/username/ is enough.

server/uploader.js
```diff
@@ -266,22 +276,28 @@ uploader.bedSort = function(prefix, bedFile, uploadPath) {
...
- * @return {boolean} File exists or not
+ * @return {boolean|uploader.Error} File exists or not
  */
 uploader.checkFinish = function(query, prefix) {
+  if (query.fileName == undefined) {
+    return {
+      error: 'fileName is undefined'
+    };
+  }
+  //TODO: convert processing progress into db.
```
yubowenok: nit. TODO format is typically

``` js
// TODO(your_name): convert ...
```

src/dialog.js
```diff
@@ -147,6 +147,7 @@ genotet.dialog.createNetwork_ = function() {
       var params = {
         type: genotet.data.ListQueryType.NETWORK
       };
+      params = {data: JSON.stringify(params)};
```
yubowenok, line 151: This seems to be heavily repeated code. Consider refactoring this to a wrapper function, or at least mark a TODO for the moment.

yubowenok:
done review @djasmine

yubowenok:
LGTM

PR #158, opened 2016-03-25
Real progress bar for data processing and move data into database

djasmine:
Add progress.php for progress monitoring.
Use data reading part for main progress proportion.
Move data name, description into database.
Change listing and get data name/description functions for database.


progress.php
```diff
@@ -0,0 +1,22 @@
...
+      break;
+    case "finish":
+      $deleteFile = array(
+        'fileName' => $fileName
+      );
+      $progress->remove($deleteFile);
+      echo 'deleted';
+      break;
+  }
+?>
```
yubowenok: EOL

progress.php
```diff
@@ -0,0 +1,22 @@
+<?php
+  $m = new MongoClient("mongodb://localhost:27017");
```
yubowenok: Single quotes are preferred in php.

server/datadb.js
```diff
@@ -0,0 +1,134 @@
...
+ */
+
+var mongodb = require('mongodb');
+var MongoClient = mongodb.MongoClient;
+var mongoUrl = 'mongodb://localhost:27017/express';
+
+var log = require('./log');
+
+/** @type {dao} */
+module.exports = datadb;
```
yubowenok: maybe dbData is better, so it's more obvious they are two words

server/datadb.js
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * @constructor
+ */
+function datadb() {}
+
+/**
+ * @const {string}
+ */
+datadb.DATA_COLLECTION = 'file';
```
yubowenok: Make private strings if possible

server/datadb.js
```diff
@@ -0,0 +1,134 @@
...
+
+  collection = db.collection(datadb.PROGRESS_COLLECTION);
+  collection.insertOne({
+    fileName: fileName,
+    user: datadb.USERNAME,
+    percentage: 0
+  }, function(err, result) {
+    if (err) {
+      log.serverLog(err);
+      callback({error: err});
```
yubowenok: Shall we return after callback with error? Otherwise the following two lines are going to be executed eventually.

server/server.js
```diff
@@ -15,6 +15,7 @@ var uploader = require('./uploader.js');
 var bed = require('./bed.js');
 var mapping = require('./mapping.js');
 var log = require('./log.js');
+var datadb = require('./datadb.js');
```
yubowenok: if the file names are two words, usually by js convention it's concatenated by dash, i.e. data-db.js. The namespace needs to be camelCase however.

server/uploader.js
```diff
@@ -58,49 +59,47 @@ uploader.query = {};
  * @param {!multer.File} file File object received from multer.
  * @param {string} prefix The destination folder to upload the file to.
  * @param {string} bigWigToWigAddr Directory of script of UCSC bigWigToWig.
- * @param {string} uploadPath Directory of temporary files.
- * @return {Object} Success or not as a JS Object.
+ * @param {!mongodb.Db} db The database object.
+ * @param {function(Object)} callback The callback function.
```
yubowenok: elaborate what an object contains as much as possible

server/uploader.js
```diff
@@ -108,20 +107,38 @@ uploader.uploadFile = function(desc, file, prefix, bigWigToWigAddr,
...
+  });
+
+  cmd = [
+    'wc',
+    '-l',
+    prefix + wigFileName
+  ].join(' ');
+  var output = childProcess.execSync(cmd).toString().split(RegExp(/\s+/));
+  var totalLine = parseInt(output[1], 10);
+  var step = parseInt(totalLine / 10, 10);
```
yubowenok: oneTenthLines or progressStep could be more obvious in what it does.

server/uploader.js
```diff
@@ -154,18 +172,30 @@ uploader.bigWigToBCWig = function(prefix, bwFile, bigWigToWigAddr, uploadPath) {
       });
       lastChrXr[chName] = xr;
       lastValue[chName] = val;
+
+      ++count;
+      if (count % step == 0) {
+        percentage += 0.1 * 80;
```
yubowenok: count / totalLine would be better I think?

server/uploader.js
```diff
@@ -196,19 +233,29 @@ uploader.bigWigToBCWig = function(prefix, bwFile, bigWigToWigAddr, uploadPath) {
...
       var segBuf = new Buffer(4 + nodes.length * uploader.DOUBLE_SIZE_);
-      segBuf.writeInt32LE(nodes.length, 0);
-      for (var i = 0, offset = 4; i < nodes.length; i++,
-        offset += uploader.DOUBLE_SIZE_) {
-        segBuf.writeDoubleLE(nodes[i], offset);
-      }
-      var fd = fs.openSync(segFile, 'w');
-      fs.writeSync(fd, segBuf, 0, offset, 0);
-      fs.closeSync(fd);
+       segBuf.writeInt32LE(nodes.length, 0);
```
yubowenok: indentations are messed up

server/uploader.js
```diff
@@ -196,19 +233,29 @@ uploader.bigWigToBCWig = function(prefix, bwFile, bigWigToWigAddr, uploadPath) {
...
+          callback(err);
+        }
+      });
     }
 
-    var fd = fs.openSync(uploadPath + bwFile + '.finish', 'w');
-    fs.writeSync(fd, 'finish');
-    fs.closeSync(fd);
+    percentage = 100;
+    datadb.updateProgress(db, bwFile, percentage, function(err) {
```
yubowenok: you can replace the percentage in the argument by 100

server/uploader.js
```diff
@@ -234,12 +292,22 @@ uploader.bedSort = function(prefix, bedFile, uploadPath) {
...
       data[parts[0]] = [];
+      chrNum++;
     }
     data[parts[0]].push({
       chrStart: parseInt(parts[1], 10),
       chrEnd: parseInt(parts[2], 10),
       name: parts[3]
     });
+    ++count;
+    if (count % step == 0) {
```
yubowenok: ditto

src/dialog.js
```diff
@@ -503,25 +503,30 @@ genotet.dialog.upload_ = function() {
...
+          type: 'finish',
+          fileName: fileName
+        };
+        $.get(genotet.data.progressURL, finishParam);
       }
-    });
-    modal.find('.progress').children('.progress-bar')
-      .css('width', widthPercent);
+    })
+      .fail(function(res) {
```
yubowenok: You can do 

``` js
.always(function(res) {
  clearInterval(interval);
});
```

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "Genotet",
   "main": "genotet.js",
   "dependencies": {
-    "bower": "^1.5.2"
+    "assert": "^1.3.0",
```
yubowenok: Are we using assert and mongodb for root level package? Aren't they for server and should be declared instead in server/package.json?

server/bed.js
```diff
@@ -5,6 +5,7 @@
 var fs = require('fs');
 
 var log = require('./log');
+var dbData = require('./dbData');
```
yubowenok: I think @Shauro is using "database". You may want to use a same namespace.

server/bed.js
```diff
@@ -77,19 +79,28 @@ bed.query.motifs = function(query, bedPath) {
...
+  bed.listBed_(db, function(data) {
+    callback(data);
+  });
 };
 // End public APIs
 
 
 /**
+ * The path name for bed data after dataPath.
+ * @private @type {string}
```
yubowenok: @private @const

server/binding.js
```diff
@@ -74,30 +76,35 @@ binding.query = {};
...
   var data = binding.getBinding_(file, query.xl, query.xr, query.numSamples);
-  data.gene = binding.getGene_(descriptionPath);
   data.chr = chr;
-  return data;
+  binding.getGene_(db, fileName, function(gene) {
+    if (gene.error) {
+      callback(gene);
+    }
+    data.gene = /** @type {string} */(gene);
+    callback(data);
```
yubowenok: If you know the type of data, then you could expand the definition of the callback param.

server/binding.js
```diff
@@ -74,30 +76,35 @@ binding.query = {};
  *   xr: (number|undefined),
  *   numSamples: (number|undefined)
  * }} query
- * @param {string} bindingPath
- * @return {binding.Histogram|binding.Error}
+ * @param {string} dataPath
+ * @param {function(*)} callback
```
yubowenok: Try to list the arguments of the callback in more detail.

server/binding.js
yubowenok: I think you did expand the arguments for other callback functions. Therefore you may want to do the same here.

server/binding.js
```diff
@@ -545,63 +560,38 @@ binding.loadHistogram_ = function(file) {
...
-  return ret;
 };
 
 /**
  * Gets gene name for a specific binding file.
- * @param {string} descriptionPath Path to the wiggle folder.
- * @return {string} the gene name.
+ * @param {!mongodb.Db} db The database object.
+ * @param {string} fileName The binding file name.
+ * @param {function(*)} callback The callback function.
```
yubowenok: ditto

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+ *   error: string
+ * }}
+ */
+dbData.Error;
+
+/**
+ * Creates a connection to the mongodb.
+ * @param {function(!mongodb.Db, Object=)} callback The callback function.
+ */
+dbData.createConnection = function(callback) {
```
yubowenok: Add a TODO. Later this shall go to server.js midware.

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+    callback(db, err);
+  });
+};
+
+/**
+ * Inserts a new file into database.
+ * @param {!mongodb.Db} db The database.
+ * @param {string} path Path to the folder of the file.
+ * @param {string} fileName File name of the file.
+ * @param {!Object} property Properties of the file.
```
yubowenok: Can you list the attributes of this Object? looks like you are inserting this to db so we must know what exactly the attributes are.

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+dbData.insertFile = function(db, path, fileName, property, callback) {
+  var collection = db.collection(dbData.DATA_COLLECTION_);
+  collection.insertOne({
+    fileName: fileName,
+    path: path,
+    property: property,
+    user: dbData.USERNAME_
+  }, function(err, result) {
+    if (err) {
+      log.serverLog(err);
```
yubowenok: I think err is a compound object. You may want to log err.message. But log(err) might also work if err has a default toString method. Could you check that if this log is behaving correctly?

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * @private @const {string}
+ */
+dbData.PROGRESS_COLLECTION_ = 'uploadProgress';
+
+/**
+ * @private @type {string}
+ */
+dbData.USERNAME_ = 'anonymous';
```
yubowenok: Make this an api, e.g.

``` js
dbData.getUsername = function() {
  return 'anonymous';
};
```

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Updates processing progress in database.
+ * @param {!mongodb.Db} db The database.
+ * @param {string} fileName File name of the file.
+ * @param {number} percentage Processing progress percentage.
+ * @param {function(Object)} callback Callback function.
+ */
+dbData.updateProgress = function(db, fileName, percentage, callback) {
+  var collection = db.collection(dbData.PROGRESS_COLLECTION_);
+  collection.updateOne({fileName: fileName},
```
yubowenok: A little weird formatting.
Suggest:

``` js
collection.updateOne(
  {fileName: fileName},
  {$set: {percentage: percentage}}, 
  function(err, result) {
    if (err) {
      log.serverLog(err);
      callback({error: err});
    }
    log.serverLog(result);
    callback({});
  }
);
```

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+ * @param {string} type The file type to list.
+ * @param {function(*)} callback
+ */
+dbData.getList = function(db, type, callback) {
+  var collection = db.collection(dbData.DATA_COLLECTION_);
+  var cursor = collection.find({
+    user: dbData.USERNAME_
+  });
+  var ret = [];
+  cursor.toArray(function(err, docs) {
```
yubowenok: Why are we using toArray rather than the each() iterator here?

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+    if (err) {
+      log.serverLog(err);
+      callback({error: err});
+    }
+    log.serverLog(docs.length);
+    if (docs.length) {
+      docs.forEach(function(doc) {
```
yubowenok: Again I think if length is zero then forEach has no effect.

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Inserts chromosomes to database for binding files.
+ * @param {!mongodb.Db} db The database object.
+ * @param {string} fileName The binding file name.
+ * @param {!Array<string>} chrs The chromosomes to be inserted.
+ * @param {function(*)} callback The callback function.
+ */
+dbData.insertBindingChrs = function(db, fileName, chrs, callback) {
+  var collection = db.collection(dbData.DATA_COLLECTION_);
+  collection.updateOne({
```
yubowenok: Suggest

``` js
collection.updateOne(
  {fileName: fileName, user: dbData.USERNAME_},
  {$set: {chrs: chrs}}, 
  function(err, result) {
    log.serverLog(result);
  }
);
```

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+    fileName: fileName,
+    user: dbData.USERNAME_
+  }, {$set: {chrs: chrs}}, function(err, result) {
+    log.serverLog(result);
```
yubowenok: Not checking err here?

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+ * @param {!mongodb.Db} db The database object.
+ * @param {string} fileName The file which the gene belongs to.
+ * @param {function(*)} callback The callback function.
+ */
+dbData.getBindingGene = function(db, fileName, callback) {
+  var collection = db.collection(dbData.DATA_COLLECTION_);
+  collection.find({
+    fileName: fileName,
+    user: dbData.USERNAME_
+  }).toArray(function(err, docs) {
```
yubowenok: I think @Shauro is using each. Please revisit the decision.

server/expression.js
```diff
@@ -90,14 +91,15 @@ expression.query = {};
...
+ * @param {string} dataPath
  * @return {expression.MatrixInfo|expression.Error}
  */
-expression.query.matrixInfo = function(query, expressionPath) {
+expression.query.matrixInfo = function(query, dataPath) {
   if (query.fileName === undefined) {
     return {error: 'fileName is undefined'};
   }
-  var file = expressionPath + query.fileName + '.data';
+  var expressionPath = dataPath + 'anonymous/' + expression.EXPRESSION_PREFIX_;
```
yubowenok: username could be get via the user api

``` js
user.getUsername();
```

There should be somewhere checking that the username retrieved is valid.

server/server.js
```diff
@@ -153,141 +128,168 @@ app.post('/genotet/log', function(req, res) {
...
+      function(err) {
+        data = err;
+      });
   });
+  res.header('Access-Control-Allow-Origin', '*');
+  if (data.error) {
+    res.status(500).json(data.error);
+  } else {
+    res.json({
+      success: true
```
yubowenok: If not 500 then it is success I think? This block of code is exactly the same as below nested "serverReponse".

server/server.js
```diff
@@ -153,141 +128,168 @@ app.post('/genotet/log', function(req, res) {
...
-      break;
-    case network.QueryType.INCREMENTAL_EDGES:
-      data = network.query.incrementalEdges(query, networkPath);
-      break;
+  dbData.createConnection(function(db, err) {
+    switch (type) {
+      // Network data queries
+      case network.QueryType.NETWORK:
+        data = network.query.network(query, dataPath);
+        serverResponse();
```
yubowenok: Avoid accessing variable outer to the scope. Prefer

``` js
serverResponse(data);
```

server/uploader.js
```diff
@@ -52,55 +53,61 @@ uploader.query = {};
...
  */
-uploader.uploadFile = function(desc, file, prefix, bigWigToWigAddr,
-                               uploadPath) {
+uploader.uploadFile = function(desc, file, prefix, bigWigToWigAddr, db,
+                               callback) {
   var fileName = file.originalname;
-  var storedName = fileName + '.data';
-  if (fs.existsSync(prefix + fileName)) {
-    fs.unlinkSync(prefix + fileName);
+  var filePath = prefix + 'anonymous/';
```
yubowenok: You may want to global search for anonymous and replace them by getUsername().

server/uploader.js
```diff
@@ -108,20 +115,38 @@ uploader.uploadFile = function(desc, file, prefix, bigWigToWigAddr,
  * @param {string} prefix Folder that contains the bw file.
  * @param {string} bwFile Name of the bigwig file (without prefix).
  * @param {string} bigWigToWigAddr The convention script path.
- * @param {string} uploadPath Directory of temporary files.
+ * @param {!mongodb.Db} db The database object.
+ * @param {function(*)} callback The callback function.
  */
-uploader.bigWigToBCWig = function(prefix, bwFile, bigWigToWigAddr, uploadPath) {
+uploader.bigWigToBCWig = function(prefix, bwFile, bigWigToWigAddr, db,
```
yubowenok: nit: bigWigToB**cw**ig

server/uploader.js
```diff
@@ -108,20 +115,38 @@ uploader.uploadFile = function(desc, file, prefix, bigWigToWigAddr,
...
+      callback(err);
+    }
+  });
+
+  cmd = [
+    'wc',
+    '-l',
+    prefix + wigFileName
+  ].join(' ');
+  var output = childProcess.execSync(cmd).toString().split(RegExp(/\s+/));
```
yubowenok, line 195: childProcess.exec() could by an async version, which is potentially better? You may mark a TODO here.

server/uploader.js
```diff
@@ -132,6 +157,7 @@ uploader.bigWigToBCWig = function(prefix, bwFile, bigWigToWigAddr, uploadPath) {
     terminal: false
   });
   var lastChrXr = {}, lastValue = {};
+  var count = 0;
```
yubowenok: lineCount

server/uploader.js
```diff
@@ -188,6 +228,13 @@ uploader.bigWigToBCWig = function(prefix, bwFile, bigWigToWigAddr, uploadPath) {
       var fd = fs.openSync(bcwigFile, 'w');
       fs.writeSync(fd, bcwigBuf, 0, uploader.ENTRY_SIZE_ * seg[chr].length, 0);
       fs.closeSync(fd);
+      percentage += 4 / chrNum;
```
yubowenok: How is 4 derived? Suggest constants.

server/uploader.js
```diff
@@ -204,11 +251,26 @@ uploader.bigWigToBCWig = function(prefix, bwFile, bigWigToWigAddr, uploadPath) {
       var fd = fs.openSync(segFile, 'w');
       fs.writeSync(fd, segBuf, 0, offset, 0);
       fs.closeSync(fd);
+      percentage += 4 / chrNum;
```
yubowenok: This part seems to be exactly the same as above. Consider creating a nested function.

server/uploader.js
```diff
@@ -204,11 +251,26 @@ uploader.bigWigToBCWig = function(prefix, bwFile, bigWigToWigAddr, uploadPath) {
...
+      dbData.updateProgress(db, bwFile, percentage, function(err) {
+        if (err.error) {
+          log.serverLog(err);
```
yubowenok, line 306: Should this log err.error?

server/uploader.js
```diff
@@ -217,16 +279,27 @@ uploader.bigWigToBCWig = function(prefix, bwFile, bigWigToWigAddr, uploadPath) {
  * Sorts bed data segments and write it into separate files.
  * @param {string} prefix Folder to the bed files.
  * @param {string} bedFile File name of the bed file.
- * @param {string} uploadPath Directory of temporary files.
+ * @param {!mongodb.Db} db The database object.
+ * @param {function(*)} callback The callback function.
  */
-uploader.bedSort = function(prefix, bedFile, uploadPath) {
-  var storedName = bedFile + '.data';
+uploader.bedSort = function(prefix, bedFile, db, callback) {
```
yubowenok, line 335: ditto changes for this function, as for bigWigToBcwig.

src/data.js
```diff
@@ -37,6 +37,12 @@ genotet.data.uploadURL;
 genotet.data.logURL;
 
 /**
+ * URL for queries for monitoring data processing.
+ * @type {string}
+ */
+genotet.data.progressURL;
```
yubowenok: Url instead of URL. For other variables in this file as well.

src/data.js
```diff
@@ -135,9 +141,12 @@ genotet.data.init = function() {
   if (window.location.protocol == 'file:') {
     // Testing environment
     genotet.data.serverURL = 'http://localhost:3000/genotet';
+    genotet.data.progressURL = 'http://localhost/genotet/progress.php';
   } else {
     genotet.data.serverURL = window.location.protocol + '//' +
       window.location.hostname + ':3000/genotet';
+    genotet.data.progressURL = window.location.protocol + '//' +
```
yubowenok: Could be more verbose as uploadProgressUrl.

test/mongo-init.js
```diff
@@ -0,0 +1,2 @@
+db.createCollection('file');
+db.createCollection('uploadProgress');
```
yubowenok: EOL

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+ * @param {!mongodb.Db} db The database.
+ * @param {string} fileName File name of the file.
+ * @param {number} percentage Processing progress percentage.
+ * @param {function(Object)} callback Callback function.
+ */
+dbData.updateProgress = function(db, fileName, percentage, callback) {
+  var collection = db.collection(dbData.PROGRESS_COLLECTION_);
+  collection.updateOne({fileName: fileName},
+    {$set: {percentage: percentage}}, function(err, result) {
+      if (err) {
```
yubowenok: @Shauro @djasmine 
err could be a compound object. Can we print err directly or we should use err.message? Please double-check.

Now that I see too many instances of 

``` js
if (err) {
  log.serverLog(err);
  callback({error: err});
}
```

Can we create a wrapper callback to capture error, so that the other callbacks does not need to worry about errors?

``` js
utils.captureErr = function(err, result, callback) {
  if (err) {
    log.serverLog(err);
    callback({error: err});
    return;
  }
  callback(result);
}
```

server/dbData.js
```diff
@@ -0,0 +1,182 @@
...
+  var collection = db.collection(dbData.DATA_COLLECTION_);
+  collection.insertOne({
+    fileName: fileName,
+    path: path,
+    property: property,
+    user: dbData.USERNAME_
+  }, function(err, result) {
+    if (err) {
+      log.serverLog(err);
+      callback({error: err});
```
yubowenok: Shouldn't we return immediately after error callback? Throughout this PR.

server/server.js
djasmine: So we are going to use that function? Should I define it outside app.get/app.post?

server/binding.js
djasmine: I am not clear about these comments, should I just add definitions for the callback function parameters?

server/dbData.js
djasmine: For this function, I just need exactly one result, but db.collection.findOne is deprecated. And Liana writes another getOne function in her branch, I will use her code after merging.

server/server.js
yubowenok: Yep I think you can define it outside app.get/post

server/binding.js
yubowenok: I mean do not define the callback to be function(*), try to list the attributes like function({...}).

server/dbData.js
yubowenok: Ack. Please mark a TODO.

server/dbData.js
djasmine: It seems we should immediately return when error occurs, then we should estimate err again outside this function.

yubowenok: You mean in the callback you need to do the following?

``` js
function callback(data) {
  if (data.error) {
    return data;
  }
  //...
}
```

yubowenok: Can we do this instead?  

``` js
utils.captureErr = function(err, result, callback) {
  if (err) {
    log.serverLog(err);
    return {error: err};
  }
  return callback(result);
}
```

djasmine: I mean when we use this function, and get error. It might be

``` javascript
utils.captureErr(err, result, function(data) {
  callback(data);
});
```

But when it errors, should we also return in outer function?

yubowenok: I see. Let me think about it.

server/bed.js
```diff
@@ -66,6 +67,7 @@ bed.query.motifs = function(query, bedPath) {
     return {error: 'chr is undefined'};
   }
   var fileName = query.fileName;
+  var bedPath = dataPath + 'anonymous/' + bed.BED_PREFIX_;
```
yubowenok: getUsername()?

server/binding.js
```diff
@@ -40,7 +41,7 @@ binding.Histogram;
 
 /**
  * @typedef {{
- *   error: string
+ *   error: (string|undefined)
```
yubowenok: Why do we need undefined here?

server/binding.js
```diff
@@ -74,30 +76,36 @@ binding.query = {};
...
-    return {error: 'fileName is undefined'};
+    callback({error: 'fileName is undefined'});
   }
   if (query.chr === undefined) {
-    return {error: 'chr is undefined'};
+    callback({error: 'chr is undefined'});
   }
   var fileName = query.fileName;
   var chr = query.chr;
+  var bindingPath = dataPath + 'anonymous/' + binding.BINDING_PREFIX_;
```
yubowenok: getUsername()?

server/database.js
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * @constructor
+ */
+function database() {}
+
+/**
+ * @private @const {string}
+ */
+database.DATA_COLLECTION_ = 'file';
```
yubowenok: This would better be an enum

``` js
/** @enum {string} */
database.Collection = {
  DATA: 'file',
  ...
}
```

server/expression.js
yubowenok: leftover? @djasmine

server/externs/mongodb.js
```diff
@@ -0,0 +1,71 @@
...
+mongodb.Db.prototype.collection.insertOne = function(document, writeConcern) {};
+
+/**
+ * @param {!Object} query
+ * @param {!Object} update
+ * @param {boolean=} upsert
+ * @param {boolean=} multi
+ * @param {Object=} writeConcern
+ */
+mongodb.Db.prototype.collection.update = function(query, update, upsert, multi, writeConcern) {};
```
yubowenok, line 55: Mark a TODO. I think @Shauro  is redefining those externs for more details.

server/network.js
```diff
@@ -222,6 +229,12 @@ network.query.allNodes = function(query, networkPath) {
 // End public APIs
 
 /**
+ * Path after data path for network files.
+ * @private @const {string}
+ */
+network.NETWORK_PREFIX_ = 'network/';
```
yubowenok: A bit redundant naming. Could be PATH_PREFIX_... Avoid the same words with namespace.

server/server.js
```diff
@@ -187,107 +156,130 @@ app.post('/genotet/upload', upload.single('file'), function(req, res) {
...
-      data = expression.query.profile(query, expressionPath);
-      break;
-    case expression.QueryType.TFA_PROFILE:
-      data = expression.query.tfaProfile(query, expressionPath);
-      break;
+  database.createConnection(function(db, err) {
+    switch (type) {
+      // Network data queries
+      case network.QueryType.NETWORK:
+        data = network.query.network(query, dataPath);
```
yubowenok: If the datas are all local scope, can we just do 

``` js
serverResponse(network.query.network(query, dataPath), res);
```

It seems that there is no need for the global (relatively) var data outside the switch().

server/uploader.js
yubowenok, line 195: mark a TODO? @djasmine

server/user.js
```diff
@@ -0,0 +1,27 @@
...
+ * Gets current username.
+ * @return {string} The username.
+ */
+user.getUsername = function() {
+  return user.USERNAME_;
+};
+
+// End Public APIs
+
+/**
```
yubowenok, line 23: Can be written in one line

``` js
/** @private @type {string} */
```

src/data.js
```diff
@@ -22,19 +22,25 @@ genotet.data = {};
...
  * @type {string}
  */
-genotet.data.logURL;
+genotet.data.logUrl;
+
+/**
+ * Url for queries for monitoring data processing.
+ * @type {string}
+ */
+genotet.data.lploadProgressUrl;
```
yubowenok: uploadProgressUrl

yubowenok: Looks like error from global replace. Be careful. With WebStorm you can actually view the list of find results before replacing them.

yubowenok:
Please edit PR title. @djasmine

yubowenok:
LGTM with minor issues.

server/uploader.js
djasmine, line 195: Marked at another childProcess.execSync()...Because we are using command line to do both bigWig to wig transfer and line count.

yubowenok, line 195: Oh ic

PR #159, opened 2016-03-30
Block front-end when server is down.

djasmine:
Block front-end when server is down.
Change readme.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,13 +34,8 @@ Installation:
 - Create and edit a server configuration file to set the data paths. The file shall be located at _genotet/server/config_.
 
     ```
-    networkPath = .../genotet_data/network/
-    bindingPath = .../genotet_data/binding/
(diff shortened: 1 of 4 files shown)


templates/server-down.html
```diff
@@ -0,0 +1,10 @@
+<div class="modal-header">
+  <h4 class="modal-title">Server not reachable</h4>
```
yubowenok: Uppercase first letters? Server Not Reachable

README.md
```diff
@@ -54,3 +49,7 @@ Installation:
     ```
     node server/server.js
     ```
+    Or
+    ```
+    bash start.sh
```
yubowenok, line 50: Maybe bash start.sh is just better and we can remove the other in Readme

yubowenok:
Sync?

yubowenok:
LGTM

PR #162, opened 2016-04-05
Optimisation for Network View

djasmine:
Iterate 100 times and stop force.
Make minimum redraw zoom scale to 40%.

### DIFF
--- a/src/components/network/network-renderer.js
+++ b/src/components/network/network-renderer.js
@@ -220,10 +220,15 @@ genotet.NetworkRenderer.prototype.zoomHandler_ = function() {
   this.canvas.selectAll('.render-group')
     .attr('transform', genotet.utils.getTransform(translate, scale));
 
-  this.zoomTranslate_ = translate;
-  this.zoomScale_ = scale;
-
-  this.drawNetwork_();
+  console.log(scale);
+  // var nodeNum = this.network.nodes.length;
+  // var normScale = nodeNum / 100;
+  if (Math.floor(scale / this.zoomScale_) > 1.2 ||
+    Math.floor(scale / this.zoomScale_) < 0.8) {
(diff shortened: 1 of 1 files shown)


src/components/network/network-renderer.js
```diff
@@ -74,6 +74,12 @@ genotet.NetworkRenderer = function(container, data) {
   this.zoomScale_ = 1.0;
   /** @private {genotet.NetworkRenderer.MouseState_} */
   this.mouseState_ = genotet.NetworkRenderer.MouseState_.NONE;
+
+  /** @private {number} */
+  this.maxTickNum_ = 100;
```
yubowenok: Please make this a constant under prototype, and do the same for minZoomRadio

src/components/network/network-renderer.js
```diff
@@ -74,6 +74,12 @@ genotet.NetworkRenderer = function(container, data) {
   this.zoomScale_ = 1.0;
   /** @private {genotet.NetworkRenderer.MouseState_} */
   this.mouseState_ = genotet.NetworkRenderer.MouseState_.NONE;
+
+  /** @private @const {number} */
+  this.maxTickNum_ = 100;
```
yubowenok: I mean move it to prototype.

``` js
genotet.NetworkRenderer.prototype.MAX_TICK_NUM_ = 100;
```

src/components/network/network-renderer.js
```diff
@@ -4,6 +4,12 @@
 
 'use strict';
 
+/** @private @const {number} */
+genotet.NetworkRenderer.prototype.MAX_TICK_NUM_ = 100;
```
yubowenok: This will not work. You need to put this after genotet.NetworkRenderer is defined.

PR #163, opened 2016-04-14
switch to jsonp to allow for session sharing across ports; refine testing db init

yubowenok:


gulp/build.js
```diff
@@ -14,8 +14,6 @@ gulp.task('build', function(cb) {
 gulp.task('build-dev', function(cb) {
   runSequence(
     'dist',
-    'compile-all',
-    'dist',
```
djasmine, line 18: Is it right to delete these?

yubowenok, line 18: It makes build-dev much faster. If we want to compile, then we could do compile-all.

djasmine:
LGTM

PR #164, opened 2016-04-16
convert all requests to request.* APIs; + batch testing script

yubowenok:
- use JSON for POST and JSONP for GET
- use request.\* APIs in all front-end requests
- fix gulp test to NOT clear data before testing
- bash test/test.sh to test everything. run server testing BEFORE front-end testing, so that the front-end may successfully retrieve data.


liana-yang:
LGTM

PR #165, opened 2016-04-19
Change the expression heatmap to canvas.

liana-yang:


yubowenok:
Performance:
- Resizing the view shall not trigger re-render. Otherwise it is impossible to resize when the matrix shows all cells.

Bugs:
- Dragging the zoom box **sometimes** does not work.  Only after **moving the mouse for some distance** does the drag box appear. The drag box is then positioned incorrectly with shifts. 
- Mousedown -> move left+up -> move right+down: The box then does not reflect the current range. That is, the mouse is inside the drag box. However, it should be the range defined by MouseDownPosition -> CurrentMousePosition.
- After zooming out and then resizing, the canvas size does not get resized correctly.
  image

liana-yang:
@yubowenok You can review this pr now.

yubowenok:
Can we prevent the canvas from being larger than the view when resizing to a smaller size? One possible solution would be to scale the canvas during the view resize, but without a redraw (Or even simpler, you can just truncate the canvas during resizing).

yubowenok:
This css might work for the purpose.

``` css
.view {
  overflow: hidden;
}
```

However, I'm not sure if there will be any side effects somewhere else.

liana-yang:
@yubowenok It works well.

yubowenok:
LGTM

PR #168, opened 2016-04-21
Identify user with session

djasmine:
Enable session to identify user.
Add batch upload python script.

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -157,55 +157,6 @@ app.use(session({
 var exonFile = dataPath + 'exons.bin';
 
 /**
- * User authenticate handler.
- */
-app.use('/genotet', function(req, res, next) {
-
-  if (req.url.substr(0, 6) == '/check') {
-    res.header('Access-Control-Allow-Origin', 'http://localhost');
-    res.jsonp({});
-  }
-  else if (req.url.substr(0, 5) == '/user') {
-    next();
-  } else {
-    if (req.session.id === undefined) {
-      res.header('Access-Control-Allow-Origin', 'http://localhost');
-      log.serverLog('user not recognized.');
-      res.jsonp({
-        error: 'user not recognized.'
-      });
-    }
-    else {
-      user.findUsername(req.session.id, function(result) {
-        if (!result.error) {
-          log.serverLog('username', result);
-          req.username = result;
-          next();
-        } else {
-          log.serverLog(result.error);
-          if (result.error == 'no user info find.') {
-            req.username = 'anonymous';
-            next();
-          } else {
-            res.header('Access-Control-Allow-Origin', 'http://localhost');
-            res.jsonp(result);
-          }
-        }
-      });
-    }
-  }
-});
-
-/**
- * @type {!Array<string>}
- */
-var allowedOrigins = [
-  // TODO(bowen): need to change cross-origin to https://localhost as well
-  'http://localhost',
-  'file://'
-];
-
-/**
  * Sends back JSON response.
  * @param {Object|user.Error|undefined} data Server response data.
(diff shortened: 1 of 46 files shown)


yubowenok:
Please sync with the With Credentials APIs

server/server.js
```diff
@@ -156,6 +157,46 @@ app.use(session({
 var exonFile = dataPath + 'exons.bin';
 
 /**
+ * User authenticate handler.
+ */
+app.use('/genotet', function(req, res, next) {
+
+  if (req.url.substr(0, 6) == '/check') {
```
yubowenok: You can use a separate get handler before this authentication handler?

```
app.get('/genotet/check', ...);
```

server/server.js
```diff
@@ -156,6 +157,46 @@ app.use(session({
...
+    res.header('Access-Control-Allow-Origin', 'http://localhost');
+    res.jsonp({});
+  }
+  else if (req.url.substr(0, 5) == '/user') {
```
yubowenok: "else if" needs to be on the previous line.

server/server.js
```diff
@@ -156,6 +157,46 @@ app.use(session({
...
+    next();
+  } else {
+    if (req.session.id === undefined) {
+      res.header('Access-Control-Allow-Origin', 'http://localhost');
```
yubowenok: Why not moving jsonResponse above and use that function instead?

server/server.js
```diff
@@ -156,6 +157,46 @@ app.use(session({
...
+      log.serverLog('user not recognized.');
+      res.jsonp({
+        error: 'user not recognized.'
+      });
+    }
+    else {
```
yubowenok: else needs to be on the previous line.

server/server.js
```diff
@@ -156,6 +157,46 @@ app.use(session({
...
+          log.serverLog('username', result);
+          req.username = result;
+          next();
+        } else {
+          log.serverLog(result.error);
+          if (result.error == 'no user info find.') {
+            req.username = 'anonymous';
+            next();
+          } else {
+            res.header('Access-Control-Allow-Origin', 'http://localhost');
```
yubowenok: You may use jsonResponse wrapper?

server/server.js
```diff
@@ -179,6 +220,7 @@ var jsonpResponse = function(data, res) {
  */
 var jsonResponse = function(data, res) {
   // TODO(bowen): need to change cross-origin to https://localhost as well
+  res.header('Access-Control-Allow-Credentials', 'true');
```
yubowenok: Please sync.

server/server.js
```diff
@@ -194,9 +236,10 @@ var jsonResponse = function(data, res) {
 /**
  * User log POST handler.
  */
-app.post('/genotet/log', function(req, res) {
+app.get('/genotet/log', function(req, res) {
```
yubowenok: Since now get/post both works with credentials, maybe we can just use post here.

server/server.js
```diff
@@ -211,55 +254,58 @@ app.post('/genotet/upload', upload.single('file'), function(req, res) {
...
+  uploader.uploadFile(body, req.file, dataPath, bigWigToWigPath, req.username,
     function(ret) {
       jsonResponse(/** @type {Object} */(ret), res);
     });
 });
 
 
-app.post('/genotet/user', function(req, res) {
-  log.serverLog('POST user');
+app.get('/genotet/user', function(req, res) {
```
yubowenok: Since now get/post both works with credentials, maybe we can just keep POST here as well?

djasmine:
You can review it again. @yubowenok

yubowenok:
LGTM

PR #169, opened 2016-04-23
Setup HTTPS server.

liana-yang:


server/server.js
```diff
@@ -415,7 +415,15 @@ MongoClient.connect(mongoUrl, function(err, mongoClient) {
...
-  var server = app.listen(3000);
-  server.setTimeout(1200000);
+  var privateKey = fs.readFileSync(privateKeyPath);
+  var certificate = fs.readFileSync(certificatePath);
+  var httpsOptions = {
+    key: privateKey,
+    cert: certificate
+  };
+  var httpsServer = https.createServer(
+    /** @type {?} */(httpsOptions),
```
yubowenok: Do not use '?'. Follow the extern defs in https.js.

yubowenok:
The build failed.

server/server.js
```diff
@@ -415,7 +415,15 @@ MongoClient.connect(mongoUrl, function(err, mongoClient) {
...
+    /** @type {?} */(app)
+  ).listen(3000);
+  /** @type {?} */(httpsServer).setTimeout(1200000);
```
yubowenok: Define httpsServer properly. Either write new externs or use existing ones.

liana-yang:
@yubowenok Test passed.

server/externs/https.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * @fileoverview Node.js express externs.
```
yubowenok: Node.js HTTPS externs.

server/server.js
```diff
@@ -157,6 +157,15 @@ app.use(session({
 var exonFile = dataPath + 'exons.bin';
 
 /**
+ * @type {!Array<string>}
+ */
+var allowedOrigins = [
+  // TODO(bowen): need to change cross-origin to https://localhost as well
```
yubowenok: You can remove this TODO now

server/server.js
```diff
@@ -252,22 +261,13 @@ app.use('/genotet', function(req, res, next) {
         next();
       } else {
         log.serverLog(result.error);
-        jsonResponse(result, req, res);
+        jsonResponse(/** @type {Object} */(result), req, res);
```
yubowenok: You defined user.findUsername to accept a callback: function(user.Error|string). That definition is probably conflicting with this one. 
https://github.com/ViDA-NYU/genotet/blob/a60493079bb4db8c15719f629c20983d3ac9ac8d/server/user.js#L262

yubowenok: In fact, that user.findUsername calls database.getOne, of which the callback is function(Object).
https://github.com/ViDA-NYU/genotet/blob/a60493079bb4db8c15719f629c20983d3ac9ac8d/server/database.js#L42

Your parsed the username in function(Object), which gives you a string. Then you use the string to callback back here. Thus it seems that result must be a string in this case.

liana-yang:
@yubowenok You can review again now.

server/server.js
```diff
@@ -157,8 +157,16 @@ app.use(session({
...
+ */
+var allowedOrigins = [
+  'https://localhost',
+  'file://'
+];
+
+/**
  * Sends back JSON response.
- * @param {Object|user.Error|undefined} data Server response data.
+ * @param {Object|user.Error|string|undefined} data Server response data.
```
yubowenok, line 169: Actually it is a little bit weird that this needs user.Error. If it needs user.Error, then it could need binding.Error/network.Error/... as well. I will leave it for your discretion.

yubowenok:
LGTM

PR #174, opened 2016-04-27
fix all tests

yubowenok:
- Improve test shell script.
- Configure tests properly to circumvent server certificate check.
- Provide mechanism for independent jasmine-node spec to share a same session, so that they appear to act as a same user.
- Fix user authentication bugs. Refine user authentication and DB codes.
- Refine qunit test structure to support async stepping.
- Add tests for user sign-up, sign-in, log-out, auto-log-in.
- Add test to sign-in on front-end before fetching test data.
- User queries validation. Check undefined username, email, etc.
- User TravisCI container-based infrastructure to speed up building and testing.


yubowenok:
Travis needs some patches. Hold on.

Issue #177, opened 2016-04-30
network label size changes upon zooming


yubowenok:

djasmine:
Fixed

Issue #176, opened 2016-04-30
show processing message when layouting network


yubowenok:
When the network is layouting, the view shows the previous network. This will be confusing as it seems like the newly set genes have no effect.

djasmine:
Fixed

PR #179, opened 2016-05-05
Fix loading info and zoom label

djasmine:
Show loading data info when ticking for new network layout #176;
Keep node label size fixed when zooming #177;
Fix minor issues for polygon selection.

### DIFF
--- a/src/components/network/network-renderer.js
+++ b/src/components/network/network-renderer.js
@@ -657,6 +657,7 @@ genotet.NetworkRenderer.prototype.zoomMode_ = function() {
  */
 genotet.NetworkRenderer.prototype.drawPath_ = function(terminator, coords,
                                                   line) {
+  this.svgPath_.select('path').remove();
   this.svgPath_.append('path')
     .attr({
       d: line(coords)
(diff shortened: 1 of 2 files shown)


src/components/network/network-renderer.js
```diff
@@ -664,6 +667,6 @@ genotet.NetworkRenderer.prototype.drawPath_ = function(terminator, coords,
         id: 'terminator',
         d: line([coords[0], coords[coords.length - 1]])
       });
-    this.svgPath_.select('path').remove();
+    this.svgPath_.selectAll('path').remove();
```
yubowenok: Why would there be multiple paths to be removed?

yubowenok:
LGTM

src/components/network/network-renderer.js
djasmine: It seems each time it appends a 'path' for one side of the polygon.

yubowenok: That's weird. Should a polygon only have a single path?

djasmine: The function is now act as that, with a new side adding into the polygon with n sides, it will draw a new path with n + 1 sides. Finally, it might be lots of overlapped paths. I wrote this part similar to one tutorial. I think a reasonable solution is to remove the previous n-side polygon and append a new one each time.

djasmine: But it is safer to remove all sides.

yubowenok: Why not just remove the n-side polygon each time, before drawing the new polygon with n+1 sides? It seems to me that any time there only needs to be a single path on the screen.

PR #183, opened 2016-05-10
Changes for shared data

djasmine:
Changed binding, mapping, expression and bed file reading corresponding to shared data.


yubowenok:
LGTM

